PR #24, opened 2019-03-29
Make BearerToken more opaque.

### LINKED ISSUE #23: BearerToken should be more opaque
BearerToken is currently a very light newtype wrapper over a String. It derefs to str, derives Debug and implements Display. This makes it too easy to accidentally leak a token by e.g. logging some structure that contains it.

We should remove the Deref and Display implementations entirely, and change the Debug implementation to just print `BearerToken(<REDACTED>)` or something like that.

sfackler:
It's very bad to accidentally log or otherwise expose a bearer token, so
it's now more opaque by removing the Display and Deref impls and
changing Debug to print `BearerToken("REDACTED")`.

Closes #23

### DIFF
--- a/conjure-object/src/bearer_token/mod.rs
+++ b/conjure-object/src/bearer_token/mod.rs
@@ -24,7 +24,7 @@ use std::str::FromStr;
 #[cfg(test)]
 mod test;
 
-// A lookoup table mapping valid characters to themselves and invalid characters to 0. We don't actually care what
+// A lookup table mapping valid characters to themselves and invalid characters to 0. We don't actually care what
 // nonzero value valid characters map to, but it's easier to read this way. There's a test making sure that the mapping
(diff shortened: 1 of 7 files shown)


conjure-object/src/bearer_token/mod.rs
```diff
@@ -18,13 +18,12 @@ use serde::ser::{Serialize, Serializer};
...
 use std::error::Error;
 use std::fmt;
-use std::ops::Deref;
 use std::str::FromStr;
 
 #[cfg(test)]
 mod test;
 
-// A lookoup table mapping valid characters to themselves and invalid characters to 0. We don't actually care what
+// A lookup table mapping valid characters to themselves and invalid characters to 0. We don't actually care what
```
bavardage: lookoup!

PR #26, opened 2019-04-10
Respect package locations

### LINKED ISSUE #16: Respect the package paths of generated types
We'll probably want to do this via a new config option that signifies the number of components to strip from the Conjure package. For example, if your Conjure definition has the package `com.palantir.myservice.foobar`, and you strip 3 components, the types would end up in the `foobar` module.

There are some potentially awkward edge cases around collisions between e.g. `com.foobar.MyType` and `com.foobar.mytype.Thing` where we'd be defining the mytype` module twice.

sfackler:
Since packages are Java-style (e.g. com.palantir.conjure.api), you can
specify a prefix to strip from the package name to avoid super deep
module hierarchies.

Closes #16

### DIFF
--- a/conjure-codegen/src/context.rs
+++ b/conjure-codegen/src/context.rs
@@ -30,14 +30,20 @@ struct TypeContext {
 pub struct Context {
     types: HashMap<TypeName, TypeContext>,
     exhaustive: bool,
+    strip_prefix: Vec<String>,
 }
 
 impl Context {
-    pub fn new(defs: &ConjureDefinition, exhaustive: bool) -> Context {
+    pub fn new(defs: &ConjureDefinition, exhaustive: bool, strip_prefix: Option<&str>) -> Context {
         let mut context = Context {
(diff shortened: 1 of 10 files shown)


conjure-codegen/src/context.rs
```diff
@@ -750,6 +752,50 @@ impl Context {
...
+        } else {
+            raw
+        }
+    }
+
+    fn raw_module_path(&self, package: &str) -> Vec<String> {
+        package.split('.').map(|s| self.ident_name(s)).collect()
+    }
+
+    fn type_path(&self, this_type: &TypeName, other_type: &TypeName) -> TokenStream {
```
dtobin: Is there a test that covers the case of types referring to types that aren't direct parents/children?

sfackler: Oh good point, I'll add one covering sibling packages.

sfackler: Updated

PR #28, opened 2019-04-15
Add error support

### LINKED ISSUE #13: Add error support
The current plan is that errors turn into normal Conjure objects, along with a union-style enum for each error namespace to allow easy deserialization of a serialized error.

This is blocked on palantir/conjure#63.

sfackler:
Errors generate the same types as a normal Conjure object, but additionally implement a separate `ErrorType` trait that provides the error's code, name, and parameter safety. You can then serialize one of these types into a Conjure-generated `SerializableError` type which corresponds to the Conjure error wire format.

Note that this follows the Conjure spec rather than the Java implementation in JSON-encoding parameter values rather than stringifying them.

Closes #13


sfackler:
This now should be ready to go modulo resolution of palantir/conjure#63 (cc @iamdanfox).

.gitattributes
```diff
@@ -1,3 +1,4 @@
 conjure-codegen/src/types/** linguist-generated=true
 conjure-codegen/src/example_types/** linguist-generated=true
+conjuer-error/src/types/** linguist-generated=true
```
dtobin: conjure

sfackler:
We're now just stringifying error parameters until we figure out what do do with the broken Java implementation.

conjure-error/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+    for (key, value) in parameters {
+        if let Some(value) = value_string(&value) {
+            builder.insert_parameters(key, value);
+        }
+    }
+
+    builder
+        .error_code(error.code())
+        .error_name(error.name())
+        .build()
```
dtobin: Where is the `error_instance_id` set?

sfackler: Fixed

conjure-error/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+        Value::I64(v) => Some(v.to_string()),
+        Value::F32(v) => Some(v.to_string()),
+        Value::F64(v) => Some(v.to_string()),
+        Value::Char(v) => Some(v.to_string()),
+        Value::String(v) => Some(v.to_string()),
+        Value::Unit => None,
+        Value::Option(Some(v)) => value_string(v),
+        Value::Option(None) => None,
+        Value::Newtype(v) => value_string(v),
+        Value::Seq(_) | Value::Map(_) | Value::Bytes(_) => None,
```
dtobin: This is different behavior than the Java implementation, but this is a reasonable compromise given that the Java implementation doesn't match the spec.

conjure-codegen/src/objects.rs
```diff
@@ -28,7 +28,7 @@ pub fn generate(ctx: &Context, def: &ObjectDefinition) -> TokenStream {
 
     quote! {
         use conjure_object::serde::{ser, de};
-        use conjure_object::serde::ser::SerializeMap as SerializeMap_;
+        use conjure_object::serde::ser::SerializeStruct as SerializeStruct_;
```
dtobin: Do we have tests that check that the ser/de still round-trips with the serialization changes?

sfackler: Yeah, we have tests in conjure-test that assert a specific JSON output.

conjure-codegen/src/errors.rs
```diff
@@ -0,0 +1,66 @@
...
+use quote::quote;
+
+use crate::context::Context;
+use crate::objects;
+use crate::types::{ErrorDefinition, ObjectDefinition};
+
+pub fn generate(ctx: &Context, def: &ErrorDefinition) -> TokenStream {
+    let object = ObjectDefinition::builder()
+        .type_name(def.error_name().clone())
+        .fields(def.safe_args().iter().chain(def.unsafe_args()).cloned())
```
bavardage: spec seems iffy on whether you can share keys between safe args and unsafe args

bavardage: ok, there doesn't seem to be any validation on this and I was able to get conjure-java to produce a non-compiling error 

```java
    public static ServiceException myError(
            Throwable cause, int param, List<String> param) {
        return new ServiceException(
                MY_ERROR,
                cause,
                SafeArg.of("param", param),
                UnsafeArg.of("param", param));
    }
```

bavardage: (so probably not a thing to worry about here, but I'll put a bug on the conjure IR project)

bavardage: https://github.com/palantir/conjure/issues/285

PR #29, opened 2019-04-17
Support error propagation

sfackler:
There are two use cases we want to handle here that weren't possible
before:

1. One service makes a request to another and it returns an error. That
    by default translates into a 500, but we want to use the same error
    instance ID to make it easier to follow the flow of the error
    through the network of services.
2. One services is acting as a proxy for another service, and does want
    propagate the exact error upstream.

To support this, the ErrorType trait has a new instance_id method which
returns the error's instance ID if it's stored internally. All of the
Conjure-generated errors simply return None (and we generate a random ID
when serializing).

The new ErrorType::with_instance_id method can be used to augment an
existing error with a specific instance ID to handle case 1.

SerializableError now implements ErrorType and overrides the instance_id
method. This handles case 2.


sfackler:
The one bit of information we lose here when propagating things is that we don't know which parameters are safe. That's not a huge deal since you can always track them down from the other service with the error instance ID.

PR #33, opened 2019-04-23
Add an Error type

sfackler:
This is intended to be the standard error passed around in networked
services. It augments a normal Rust error with information about how it
should be reported to a client, extra parameters providing context, and
backtraces.

### DIFF
--- a/conjure-error/src/error.rs
+++ b/conjure-error/src/error.rs
@@ -1,9 +1,10 @@
-use backtrace::Backtrace;
 use conjure_object::Value;
+use parking_lot::Mutex;
 use serde::Serialize;
 use std::borrow::Cow;
 use std::collections::hash_map::{self, HashMap};
 use std::error;
+use std::fmt;
 use std::ops::Index;
 use std::time::Duration;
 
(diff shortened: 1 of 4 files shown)


conjure-error/src/error.rs
```diff
@@ -0,0 +1,380 @@
...
+    /// Returns the amount of time the client should wait before retrying, if provided.
+    #[inline]
+    pub fn duration(&self) -> Option<Duration> {
+        self.duration
+    }
+}
+
+/// Information about an unavailable error.
+#[derive(Debug)]
+pub struct UnavailableError(());
```
jonsyu1: Why is this a tuple struct instead of a unit struct? (i.e. `pub struct UnavailableError;`)

conjure-error/src/error.rs
```diff
@@ -0,0 +1,380 @@
...
+///
+/// An error consists of several components:
+///
+/// * The cause of the error, represented as a type implementing the Rust `Error` trait. The cause can either be
+///     declared safe or unsafe to log.
+/// * The error's kind, indicating how the service should handle the error e.g. in a response to a client.
+/// * Backtraces, including one taken at the time the error was created.
+/// * Parameters adding extra context about the error. They can be declared either safe or unsafe to log.
+///
+/// Note that this type does *not* implement the standard library's `Error` trait.
```
jonsyu1: Why doesn't this `Error` implement the standard library's `Error` trait?

conjure-error/src/error.rs
```diff
@@ -0,0 +1,380 @@
...
+        error: SerializableError,
+        safe_args: &[&str],
+    ) -> Error {
+        let mut safe_params = HashMap::new();
+        let mut unsafe_params = HashMap::new();
+
+        for (key, value) in error.parameters() {
+            let key = Cow::Owned(key.clone());
+            let value = Value::String(value.clone());
+            if safe_args.contains(&&*key) {
```
jonsyu1: Can this be simplified to `&key`?

conjure-error/src/error.rs
sfackler: We don't want to allow people to construct the type.

conjure-error/src/error.rs
jonsyu1: Ah, we're relying on `Cow`'s implementation of `Deref`.

conjure-error/src/error.rs
sfackler: It's not intended to be used as one - once you starting using `Error` we generally expect the remainder of the call chain to deal with that type specifically.

conjure-error/src/error.rs
sfackler: No - we need to go from `&Cow<str>` to `&str`.

PR #34, opened 2019-04-25
Service client generation

sfackler:
We want to enable code generation to evolve separately from the HTTP
client/server logic, so the conjure-http crate just contains a tiny shim
interface that we generate code targetting and the real clients/servers
work with generically.

In the case of clients, we generate a struct which wraps a generic
`Client` implementation and has methods per endpoint that handle the
serialization and deserialization logic.

### DIFF
--- a/conjure-codegen/src/context.rs
+++ b/conjure-codegen/src/context.rs
@@ -631,6 +631,95 @@ impl Context {
         }
     }
 
+    pub fn is_binary(&self, def: &Type) -> bool {
+        match def {
+            Type::Primitive(PrimitiveType::Binary) => true,
+            Type::Primitive(_)
+            | Type::Optional(_)
+            | Type::List(_)
+            | Type::Set(_)
+            | Type::Map(_) => false,
+            Type::Reference(def) => self.is_binary_ref(def),
+            Type::External(def) => self.is_binary(def.fallback()),
+        }
+    }
+
+    fn is_binary_ref(&self, name: &TypeName) -> bool {
+        let ctx = &self.types[name];
+
+        match &ctx.def {
+            TypeDefinition::Alias(def) => self.is_binary(def.alias()),
+            TypeDefinition::Enum(_) | TypeDefinition::Object(_) | TypeDefinition::Union(_) => false,
+        }
+    }
+
+    pub fn is_plain(&self, def: &Type) -> bool {
+        match def {
+            Type::Primitive(primitive) => match primitive {
+                PrimitiveType::String
+                | PrimitiveType::Datetime
+                | PrimitiveType::Integer
+                | PrimitiveType::Double
+                | PrimitiveType::Safelong
+                | PrimitiveType::Binary
+                | PrimitiveType::Boolean
+                | PrimitiveType::Uuid
+                | PrimitiveType::Rid
+                | PrimitiveType::Bearertoken => true,
+                PrimitiveType::Any => false,
+            },
+            Type::Optional(_) | Type::List(_) | Type::Set(_) | Type::Map(_) => false,
+            Type::Reference(def) => self.is_plain_ref(def),
+            Type::External(def) => self.is_binary(def.fallback()),
+        }
+    }
+
+    fn is_plain_ref(&self, name: &TypeName) -> bool {
+        let ctx = &self.types[name];
+
+        match &ctx.def {
+            TypeDefinition::Alias(def) => self.is_plain(def.alias()),
+            TypeDefinition::Enum(_) => true,
+            TypeDefinition::Object(_) | TypeDefinition::Union(_) => false,
+        }
+    }
+
+    pub fn is_single_value(&self, def: &Type) -> bool {
+        match def {
+            Type::Primitive(_) => true,
+            Type::Optional(_) | Type::List(_) | Type::Set(_) | Type::Map(_) => false,
+            Type::Reference(def) => self.is_single_value_ref(def),
+            Type::External(def) => self.is_single_value(def.fallback()),
+        }
+    }
+
+    fn is_single_value_ref(&self, name: &TypeName) -> bool {
+        let ctx = &self.types[name];
+
+        match &ctx.def {
+            TypeDefinition::Alias(def) => self.is_single_value(def.alias()),
+            TypeDefinition::Enum(_) | TypeDefinition::Object(_) | TypeDefinition::Union(_) => true,
+        }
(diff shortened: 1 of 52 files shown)


sfackler:
I still need to finish writing tests and documentation, but the raw code generation's there.

conjure-codegen/src/context.rs
```diff
@@ -631,6 +631,95 @@ impl Context {
...
+                | PrimitiveType::Binary
+                | PrimitiveType::Boolean
+                | PrimitiveType::Uuid
+                | PrimitiveType::Rid
+                | PrimitiveType::Bearertoken => true,
+                PrimitiveType::Any => false,
+            },
+            Type::Optional(_) | Type::List(_) | Type::Set(_) | Type::Map(_) => false,
+            Type::Reference(def) => self.is_plain_ref(def),
+            Type::External(def) => self.is_binary(def.fallback()),
```
bavardage: do you mean `is_plain` here?

conjure-codegen/src/context.rs
```diff
@@ -631,6 +631,95 @@ impl Context {
...
+            Type::External(def) => self.is_single_value(def.fallback()),
+        }
+    }
+
+    fn is_single_value_ref(&self, name: &TypeName) -> bool {
+        let ctx = &self.types[name];
+
+        match &ctx.def {
+            TypeDefinition::Alias(def) => self.is_single_value(def.alias()),
+            TypeDefinition::Enum(_) | TypeDefinition::Object(_) | TypeDefinition::Union(_) => true,
```
bavardage: wait why are these true?

what is english definition of a 'single value'?

conjure-http/src/client.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// The client's response body type.
+    type ResponseBody: Read;
+
+    /// Makes an HTTP request.
+    ///
+    /// The request's URI will be absolute-form, and it is the responsibility of the client to add the authority and
+    /// any extra context path required. The request body will be unencoded, and the request will not include a
+    /// `Content-Length` header.
+    ///
+    /// A response must only be returned if it has a 2xx status code. The client is responsible for handling all other
```
bavardage: not sure if worth mentioning 204s here

conjure-codegen/src/context.rs
sfackler: Here, "single value" is in contrast to "iterable over multiple values". Maybe this should be inverted to `is_iterable`?

conjure-codegen/src/context.rs
sfackler: Oops, fixed.

conjure-codegen/src/context.rs
bavardage: might be clearer - alternatively just brief description of what this is (thought might be super obvious from context later on)

conjure-http/src/client.rs
```diff
@@ -0,0 +1,76 @@
...
+    T: AsRef<[u8]>,
+{
+    fn write_body(&mut self, w: &mut dyn Write) -> Result<(), Error> {
+        let buf = &self.get_ref().as_ref()[self.position() as usize..];
+        w.write_all(buf).map_err(Error::internal_safe)?;
+        self.set_position(self.get_ref().as_ref().len() as u64);
+        Ok(())
+    }
+
+    fn reset(&mut self) -> bool {
```
bavardage: problematic that this sets to 0 whereas write starts from whatever self.postition() is?

sorry - don't have alternate suggestion, just weird that a 'retry' might end up writing more the second time around

sfackler: Yeah, this implementation has to assume we're starting at the beginning of the cursor.

We could split out a separate `IntoWriteBody` trait and have methods take those, so you'd pass in a Vec into the method, which then turns in a Cursor<Vec>

bavardage:
pretty squirrelly! but the resulting generated code looks good and is easy to read

conjure-http/src/client.rs
sfackler: Updated - should be easier to use.

I realized there's no reason to use a `Cursor` here in the first place, anyway!

conjure-http/src/client.rs
sfackler: I don't think the client needs to do anything special there.

bavardage: hm, so there's some specific stuff in the wire spec around 204s - but not sure how much validation should happen on the client side vs server side for that.

If result type is optional<T> and server returns empty 200, that goes against
> the de-aliased return type is optional<T> and the value is not present: servers must send 204 No Content.

but not sure if the client needs to care

https://palantir.github.io/conjure/#/docs/spec/wire?id=_31-status-codes

sfackler: The client itself doesn't have the context to know if a 204 is ok or not. It is responsible for knowing that the HTTP response is well formed (i.e. a 204 doesn't have a body), but IMO that's implied by it being an HTTP client in the first place.

sfackler:
This should be ready to go.

conjure-codegen/src/example_types/another/test_service.rs
```diff
@@ -0,0 +1,770 @@
...
+        optional_end: Option<&conjure_object::ResourceIdentifier>,
+    ) -> Result<i32, conjure_http::private::Error> {
+        let mut request_ =
+            conjure_http::private::http::Request::new(conjure_http::client::Body::Fixed(
+                conjure_http::private::json::to_vec(&query)
+                    .map_err(conjure_http::private::Error::internal)?,
+            ));
+        *request_.method_mut() = conjure_http::private::http::Method::POST;
+        let mut path_ = "/catalog/test-query-params".to_string();
+        path_.push_str("?different=");
```
bavardage: so this is good (and let's continue with this approach) but reading this method made me wonder whether we could instead just put them into the request and have the client library take care of putting them in the url?

fwiw we take this approach with python generator - just passing query params to requests via `params` argument: 
https://2.python-requests.org/en/master/user/quickstart/#passing-parameters-in-urls and https://github.com/palantir/conjure-python/blob/develop/conjure-python-core/src/main/java/com/palantir/conjure/python/poet/PythonEndpointDefinition.java#L133

sfackler: We have to generate the path anyway. I think ideally the http crate would gain a `UriBuilder` type that we could use here to simplify things.

PR #36, opened 2019-05-01
Push body serialization into client

sfackler:
This allows the client to control how e.g. serializable requests are
encoded.


sfackler:
This also removes the IntoWriteBody trait - it was added because I thought a WriteBody for &[u8] needed to be a Cursor, but that's not the case. There are some types that could need some extra state, but they're rare enough that it isn't worth complicating the API.

PR #41, opened 2019-05-05
Server generation support

sfackler:
The server interface for a service consists of a trait implemented to
handle the actual logic of the service, and a resource object which
wraps a trait implementation and handles the HTTP serialization and
deserialization glue.

### DIFF
--- a/conjure-http/src/server.rs
+++ b/conjure-http/src/server.rs
@@ -210,7 +210,7 @@ where
     const NAME: &'static str;
 
     /// Returns the resource's HTTP endpoints.
-    // ideally this would be a &'static [Endpoint] once const fns become more powerful
+    // FIXME ideally this would be a &'static [Endpoint] once const fns become more powerful
     fn endpoints() -> Vec<Endpoint<Self, B, R>>
     where
         B: RequestBody,
--- a/conjure-http/src/server.rs
+++ b/conjure-http/src/server.rs
@@ -1,0 +1,310 @@
+// Copyright 2019 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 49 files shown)


conjure-http/src/server.rs
```diff
@@ -0,0 +1,310 @@
...
+pub trait Resource<B, R>: Sized
+where
+    B: RequestBody,
+    R: VisitResponse,
+{
+    /// The resource's name.
+    const NAME: &'static str;
+
+    /// Returns the resource's HTTP endpoints.
+    // ideally this would be a &'static [Endpoint] once const fns become more powerful
```
jonsyu1: Is there an issue we can link here?

sfackler: Unclear: https://github.com/rust-lang/rust/issues/57563#issuecomment-489681952

conjure-object/src/safe_long.rs
```diff
@@ -132,3 +147,24 @@ impl fmt::Display for BoundsError {
...
+enum ParseErrorInner {
+    Parse(ParseIntError),
+    Bounds(BoundsError),
+}
+
+/// The error returned after failing to parse a string into a `SafeLong`.
+#[derive(Debug, Clone)]
+pub struct ParseError(ParseErrorInner);
+
+impl fmt::Display for ParseError {
```
jonsyu1: Do you have thoughts on [enum-display-derive](https://github.com/ihrwein/enum-display-derive)? I'm fine with this explicit implementation but I'm curious if there is an explicit reason not to use this crate.

sfackler: Doesn't seem worth it to pull in an extra dependency for one type.

PR #46, opened 2019-05-09
Parameterize binary body writers

sfackler:
Previously, `WriteBody` implementations just took `&mut dyn Write`, but this allows client and server implementations to pick their writer type explicitly.

### DIFF
--- a/conjure-test/build.rs
+++ b/conjure-test/build.rs
@@ -7,7 +7,7 @@ fn main() {
 
     println!("cargo:rerun-if-changed={}", input);
     conjure_codegen::Config::new()
-        .run_rustfmt(false)
+        // .run_rustfmt(false)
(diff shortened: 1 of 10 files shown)


conjure-test/build.rs
```diff
@@ -7,7 +7,6 @@ fn main() {
 
     println!("cargo:rerun-if-changed={}", input);
     conjure_codegen::Config::new()
-        .run_rustfmt(false)
```
jonsyu1: Out of curiosity, why don't we need this anymore?

sfackler: Disabling rustfmt makes the test build a tiny bit slower, but it makes compilation errors totally unreadable when you're messing with the codegen since it's all on one huge line before reformatting.

jonsyu1: a tiny bit faster*

PR #50, opened 2019-06-03
Record endpoint deprecation

sfackler:


bavardage:
sure - but who is going to be using this new `deprecated()` method on endpoint?

sfackler:
Server-side tracking of calls to deprecated endpoints.

PR #54, opened 2019-06-04
Add TryFrom impls for SafeLong

sfackler:


conjure-object/src/safe_long.rs
```diff
@@ -136,6 +137,42 @@ macro_rules! impl_into {
...
+                #[inline]
+                fn try_from(n: SafeLong) -> Result<$t, TryFromIntError> {
+                    n.0.try_into()
+                }
+            }
+        )*
+    };
+}
+
+impl_try_into!(u8, i8, u16, i16, u32, i32, u64, u128);
```
jonsyu1: Is there a reason we implement `TryInto` for more types than `TryFrom`?

sfackler: We never explicitly implement TryInto, but the two macros generate implementations that go in different directions - impl_try_from constructs a Safelong and impl_try_into constructs the other type.

PR #64, opened 2019-08-30
Parse unions without a "type" field

sfackler:
palantir/conjure#111

### DIFF
--- a/conjure-test/src/test/objects.rs
+++ b/conjure-test/src/test/objects.rs
@@ -129,6 +129,22 @@ fn unions() {
 }
 
 #[test]
+fn typeless_unions() {
+    test_de(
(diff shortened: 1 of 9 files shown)


conjure-test/src/test/objects.rs
```diff
@@ -128,6 +128,16 @@ fn unions() {
     test_ser(&unknown_value, unknown_json);
 }
 
+#[test]
+fn typeless_unions() {
```
bavardage: worth testing the case of deserializing empty optional? (one of the edge cases described in  https://github.com/palantir/conjure/issues/111)

bavardage:
lgtm, but I guess hold off on merging until https://github.com/palantir/conjure/issues/111 has been resolved definitively

PR #65, opened 2019-09-03
Support generation of full crates

sfackler:
## Before this PR
Previously, conjure-rust would only generate a single Rust module.

## After this PR
==COMMIT_MSG==
`conjure-rust` and `conjure-codegen` can now optionally generate full crates.
==COMMIT_MSG==

We'll use this for more straightforward gradle-conjure integration.


conjure-codegen/src/lib.rs
```diff
@@ -480,7 +568,15 @@ impl ModuleTrie {
         Ok(())
     }
 
-    fn create_root_module(&self) -> TokenStream {
+    fn create_root_module(&self, lib_root: bool) -> TokenStream {
+        let attrs = if lib_root {
+            quote! {
+                #![allow(warnings)]
```
jonsyu1: Why do we want to allow warnings if `lib_root`?

sfackler: Since we're autogenerating code it can end up triggering things like clippy, and it's easier to just turn that off rather than trying to make everything way smarter on the generation side.

jonsyu1: `create_root_module` shouldn't need to know about `lib_root`. How about defining a `fn allow_warnings() -> TokenStream` that we can `extend` with the `TokenStream` returned by `create_root_module`? This moves `lib_root` to the callers and simplifies `create_root_module`.

sfackler: It seems to make sense to me that the thing that's responsible for knowing how to create the root module for a package would want to know what kind of root it is.

jonsyu1: Sure, no strong preference. I prefer separate methods `create_root_lib` and `create_root_module` if we're making the distinction at this level.

PR #70, opened 2019-09-06
Accept productName/productVersion flags

sfackler:
## Before this PR
This generator couldn't be used with the Gradle conjure-local plugin because it always passes `--productName` and `--productVersion` flags.

## After this PR
==COMMIT_MSG==
The generator now accepts `--productName` and `--productVersion` flags as aliases for `--crateName` and `--crateVersion` respectively.
==COMMIT_MSG==


ferozco:
👍

PR #85, opened 2019-10-14
Support asynchronous clients and servers

sfackler:
Thanks to async/await the implementation and API of async clients is
pretty close to the API of the sync clients! This means that almost all
of the codegen is shared, with a flag passed around to make the few
changes we need here and there.

### DIFF
--- a/conjure-http/src/client.rs
+++ b/conjure-http/src/client.rs
@@ -22,8 +22,6 @@ use std::error;
 use std::future::Future;
 use std::io::Write;
 use std::pin::Pin;
-use tokio_io::{AsyncWrite, AsyncWriteExt};
-
 use crate::{PathParams, QueryParams};
 
 /// A trait implemented by HTTP client implementations.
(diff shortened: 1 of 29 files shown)


sfackler:
Yep

sfackler:
Added support for async servers as well.

conjure-http/src/client.rs
```diff
@@ -53,6 +55,37 @@ pub trait Client {
         U: VisitResponse<Self::BinaryBody>;
 }
 
+/// A trait implemented by async HTTP client implementations.
+pub trait AsyncClient {
```
sfackler: There's a rustc bug that prevents us from using `#[async_trait]` for this, but that's not a huge deal since we don't expect many things to actually have to implement this. https://github.com/rust-lang/rust/issues/63033

bavardage: same thing true for https://github.com/palantir/conjure-rust/pull/85/files#diff-adeb12920f549fa245147606171e1adeR45 ?

sfackler: All implementations of that should be generated by conjure-codegen, so there's no real need to make it ergonomic.

PR #98, opened 2020-02-05
Don't fail on IR field additions

sfackler:
cc palantir/conjure#494

## Before this PR
We used `server_from_str` when deserializing IR files, which fails on unknown fields.

## After this PR
==COMMIT_MSG==
`conjure-codegen` no longer fails to deserialize IR files with unknown fields.
==COMMIT_MSG==


carterkozak:
Thanks! :+1:

PR #100, opened 2020-02-12
Add a custom Any type

### LINKED ISSUE #4: Using serde-value::Value for Any misses float/bytes special casing
We presumably want deserialization from our Any equivalent to a concrete type to behave like deserialization directly from JSON to that type, but we currently will miss our special case handling of base64 bytes and non-finite floats.

We'll probably need to stop using serde-value and make our own version of the same thing that handles the special cases.

sfackler:
## Before this PR
We previously used serde_value::Value as our `any` type, but this failed to respect some of the special casing required around things like non-finite floats and binary data.

Closes #4

## After this PR
==COMMIT_MSG==
`serde_value::Value` has been replaced by a new `conjure_object::Any` type which correctly respects Conjure's serialization requirements.
==COMMIT_MSG==


conjure-object/src/any/ser.rs
```diff
@@ -0,0 +1,613 @@
...
+        _: u32,
+        variant: &'static str,
+        value: &T,
+    ) -> Result<Self::Ok, Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        let mut map = BTreeMap::new();
+        let value = value.serialize(self)?;
+        map.insert(variant.to_string(), value);
```
dtobin: Do we also need the `"type": variant` entry in the map?

sfackler: This is encoding the default serde_json enum format rather than the conjure enum format.

PR #113, opened 2020-09-02
Fix parsing of boolean options

### LINKED ISSUE #68: --exhaustive=false enables the exhaustive flag
This doesn't conform to https://github.com/palantir/conjure/blob/master/docs/rfc/002-contract-for-conjure-generators.md#options.

jdno:
## Before this PR

The CLI flag `exhaustive` is set to `true` when an argument is provided, even if the argument is `=false`. This doesn't conform with RFC-002.

The behavior can be fixed by providing a custom parser for the option, as suggested in https://github.com/TeXitoi/structopt/issues/212#issuecomment-561093886.

## After this PR

==COMMIT_MSG==
Align parsing of boolean flags with RFC 002
==COMMIT_MSG==

Closes #68

### DIFF
--- a/conjure-rust/src/main.rs
+++ b/conjure-rust/src/main.rs
@@ -33,7 +33,7 @@ enum Opts {
 // FIXME move aliases over to the standard names
 #[derive(StructOpt)]
 struct Args {
-    #[structopt(long = "exhaustive")]
+    #[structopt(long = "exhaustive", parse(try_from_str))]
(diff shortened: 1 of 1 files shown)


jdno:
Oh I didn't test what happens when no option is provided. 😬 It works when `default_value` is provided, but it's a bit ugly...

jdno:
Ok this doesn't work as expected. Let me go back to the drawing board...

PR #125, opened 2021-01-28
Fix windows dist build

sfackler:
The old installer we were using for the Windows dist was moved/removed, so we should just go the actually supported route and install via rustup.

### DIFF
--- a/conjure-object/Cargo.toml
+++ b/conjure-object/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conjure-object"
-version = "0.7.3"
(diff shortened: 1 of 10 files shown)


conjure-object/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "conjure-object"
-version = "0.7.3"
+version = "0.7.4"
```
bavardage, line 3: heh would be nice if autorelease actually took care of this stuff...

bavardage (approved):

PR #128, opened 2021-04-04
Switch publishing over to maven central

sfackler:
Publishing to Maven Central is significantly more complex than to bintray (there's GPG involved, stage/commit logic, etc), so I've just added a tiny Gradle layer so we can use our standard publishing plugin to handle all that for us.


sfackler:
Ping

jonsyu1 (approved):

PR #131, opened 2021-04-18
Implement staged builders

### LINKED ISSUE #127: Add the option to create staged builders
Following in the footsteps of https://github.com/palantir/conjure-java/pull/1243.

sfackler:
The approach here is to parameterize the builder type with a "stage" type parameter, and each required field advances that parameter to the next stage. Optional fields are settable in the terminal stage. To support modifying a value by converting it back into a builder, the setters for required fields are also callable on the terminal stage by making the methods generic over a per-field trait implemented by their respective stage type and the terminal stage type. We could make required fields settable on any stage after the first one that adds them, but that seems pretty unnecessary and introduces O(n^2) trait implementations which seems non-ideal.

Closes #127


conjure-codegen/src/objects/builder/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+                args: vec![SetterArg {
+                    name: quote!(#name),
+                    type_: quote!(T),
+                }],
+                where_: quote!(where T: #argument_bound),
+                op: SetterOp::Assign { rhs: assign_rhs },
+                name,
+            }]
+        }
+        SetterBounds::Collection {
```
jonsyu1, line 95: Factor out the business logic for this and other longer match cases.

jonsyu1 (approved):
Changes look good. I left a nitpick on code organization.

conjure-codegen/src/objects/builder/mod.rs
sfackler: Refactored things a bit to clean up.

PR #140, opened 2021-06-24
Bump the circle mac image

sfackler:
Support for 10.1.0 is being removed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
 
   dist-macos:
     macos:
-      xcode: 10.3.0
+      xcode: 12.5.0
     environment:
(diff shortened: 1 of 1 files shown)


bluekeyes (approved):
This seems fine for now, but do you know how hard it would be to move to Xcode 11 or 12? It looks like Circle will remove all Xcode 10 images except for the final release later this year once Xcode 13 is out: https://circleci.com/docs/2.0/xcode-policy/

sfackler:
Oh yeah, we can just bump up to the newest now if you'd like.

bluekeyes (approved):
Thanks, using latest should avoid having to do this upgrade again in a few months if it works.

PR #145, opened 2021-09-22
Overhaul the client API

### LINKED ISSUE #123: Simplify the client interface
The client interface currently forces a lot of the work on the HTTP client implementation - assembling the path, encoding the request, setting headers, decoding the response, etc, and the interface is a bit complex, with multiple layers of visitors and things.

We should simplify this such that the interface is just a function from `http::Request` to `http::Response`. The generated code would then be responsible for building the path, setting encoding headers, encoding the request, and decoding the response. We should be able to keep the generated code size down the same way we do now by calling out to "private" utility functions.

The only janky bit is that we'd probably need to guarantee that the request will contain a special value in the extensions containing information about the endpoint for logging and whatever else.

sfackler:
## Before this PR
The existing client interface is both complex and forced client implementations to handle a bunch of Conjure-specified logic.

## After this PR

==COMMIT_MSG==
Redesigned the Conjure client interface.
==COMMIT_MSG==

The new interface works at a much more general level, by exchanging `http::Request` and `http::Response` objects with the client implementation. The generated Conjure code is responsible for setting up the entire request by handling body serialization, path construction, etc.

Additionally, the new interface is more flexible - an HTTP client could use the Conjure client trait as its main interface and still leave consumers the flexibility to interact with non-Conjure clients.

This generally mirrors the approach we use in the Java Dialogue implementation, with the one exception that the body serde logic is not pluggable - we specifically work with only JSON. It would be possible to add the ability to customize that in the future but would require adding a second type parameter to generated client objects.

I'm working on a similar change on the server side, moving to an interface that looks more like our Java Undertow implementation and exchanges `http::Request` and `http::Response` objects.

Closes #123


sfackler:
The diff is not actually that huge - most of it is just regenerating a bunch of checked in Conjure code.

jonsyu1 (approved):

PR #146, opened 2021-10-03
Overhaul the server API

### LINKED ISSUE #122: Parse path templates in conjure-codegen
We currently provide the raw path template directly to the underlying client and server backends directly, and rely on them to parse it to perform any required parameter extraction. This results in duplicated code and the risk of divergence in how those parsers are implemented. We should instead parse the template during codegen and output a structured representation in the generated code. For the client side, we could just build the URI directly in the generated code, and in the server side we could potentially provide a list of components or something like that.

sfackler:
Following in the steps of #145, this PR redesigns the server API to move as much Conjure-specific behavior as possible into the generated code. The interface exposed to server implementations is simply a map from `Request` to `Response`.

One downside of the current interface is that you basically need an adapter layer to map `Result<Response, Error>` to `Response` since otherwise you can't just use `?` to propagate errors in a handler. Part of the awkwardness here is that we want `SafeParams` to be filled out even if the request fails. A possible alternate approach could be:

```rust
pub trait Endpoint<I, O>: EndpointMetadata {
    fn handle(&self, req: Request<I>, safe_params: &mut SafeParams) -> Result<Response<ResponseBody<O>, Error>;
}
```

It is unfortunate that we need to add this extra argument, but seems less bad than having some kind of interior-mutable extension in the request that you'd poke the safe params into. This also makes the server implementation responsible for encoding error responses, but we could provide a public API for that so you could just do something like

```rust
let mut safe_params = SafeParams::new();
let response = handler.handle(req, &mut safe_params)
    .unwrap_or_else(conjure_http::server::encode_error_response);
```

Like the previous PR, most of the line count is just regenerating the checked in code.

Closes #122


sfackler:
Flagging do-not-merge for now so I can check what the integration into our server backend will look like. Should be reviewable though.

conjure-codegen/src/http_paths.rs
```diff
@@ -0,0 +1,25 @@
+pub enum PathSegment<'a> {
+    Literal(&'a str),
+    Parameter {
+        name: &'a str,
+        regex: Option<&'a str>,
```
sfackler, line 5: This is a sneaky hidden feature in Conjure - I think the only variant we expect to actually see in practice is a regex of `.*` on the last segment to allow matches of multi-segment URLs.

conjure-codegen/src/servers.rs
```diff
@@ -235,442 +222,482 @@ enum ReturnType<'a> {
...
         }
-        Style::Sync => quote!(),
-    };
+    });
 
-    let extract_path_params = extract_path_params(ctx, endpoint, &path_params);
-    if extract_path_params.is_empty() {
-        path_params = quote!(_);
-    }
+    let template = &***endpoint.http_path();
```
jonsyu1, line 364: Why do we need to dereference this 3 times? The first dereference (`HttpPath::deref`) gives us a `String`. The second dereference (`String::deref`) gives us a `str`. It looks like we want `template` to be a `&str`, so I expected this to be `&**endpoint.http_path()`.

sfackler: The extra `*` is to go from `&HttpPath` to `HttpPath` I believe.

jonsyu1: Ah that would make sense. Thanks.

jonsyu1 (approved):
Nice, the generated code is more readable.

PR #148, opened 2021-10-07
Include recommended product dependencies in crate metatata

sfackler:
The gradle-conjure plugin adds declared service dependencies as an extension in the generated Conjure IR. This PR causes conjure-rust to forward that data into the Cargo manifest metadata section where it can be picked up by downstream build tooling to automatically pick up service dependencies based on Conjure APIs the crate depends on.

### DIFF
--- a/changelog/@unreleased/pr-148.v2.yml
+++ b/changelog/@unreleased/pr-148.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: Recommended service dependency information is encoded in the generated
+    crate's manifest metadata section.
+  links:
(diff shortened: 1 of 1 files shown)


sfackler:
Example Cargo.toml from an IR file that includes a product dep:

```toml
[package]
name = 'aip-api'
version = '0.1.0'
edition = '2018'
[[package.metadata.sls.recommended-product-dependencies]]
maximum-version = '1.x.x'
minimum-version = '1.80.0'
optional = false
product-group = 'com.palantir.myservice'
product-name = 'myservice'

[dependencies]
conjure-error = '0.7.4'
conjure-http = '0.7.4'
conjure-object = '0.7.4'
```

PR #190, opened 2022-04-24
Migrate to Clap v3

sfackler:


conjure-rust/src/main.rs
```diff
@@ -13,63 +13,61 @@
...
     use_staged_builders: bool,
-    #[structopt(long = "strip-prefix", value_name = "prefix", alias = "stripPrefix")]
+    #[clap(long = "stripPrefix", value_name = "prefix", alias = "strip-prefix")]
     /// Strip a prefix from types's package paths
     strip_prefix: Option<String>,
     /// The name of the generated crate
-    #[structopt(
-        long = "crate-name",
+    #[clap(
+        long = "productName",
```
jonsyu1, line 45: Why are we changing `long` from `crate-name` to `productName`? Same question below for version.

sfackler: We originally used `kebab-case` names for the command line options, but the Conjure CLI spec mandates `camelCase` (I think to more easily configure things in gradle-conjure?). We've supported both for a long time via aliases, but this just switches the defaults around so the `kebab-case` names are the aliases while the `camelCase` names are the normal ones that show up in `--help` etc.

jonsyu1 (approved):

PR #199, opened 2022-05-25
Use prettyplease for code formatting

### LINKED ISSUE #197: Use prettyfmt instead of shelling out to rustfmt in codegen
https://github.com/dtolnay/prettyplease

sfackler:
This avoids having a dependency on rustfmt being installed in the environment, and deals a bit better with autogenerated code in some contexts.

The vast majority of the change is just style changes to the generated code we check in!

Closes #197

### DIFF
--- a/conjure-codegen/src/example_types/another/test_service.rs
+++ b/conjure-codegen/src/example_types/another/test_service.rs
@@ -1,4 +1,4 @@
-#[doc = "A Markdown description of the service."]
+///A Markdown description of the service.
 #[derive(Clone, Debug)]
 pub struct TestServiceAsyncClient<T>(T);
 impl<T> conjure_http::client::AsyncService<T> for TestServiceAsyncClient<T>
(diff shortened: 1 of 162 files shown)


conjure-codegen/src/example_types/another/test_service.rs
```diff
@@ -13,14 +13,17 @@ impl<T> TestServiceAsyncClient<T>
...
     ) -> Result<
-        std::collections::BTreeMap<String, super::super::product::datasets::BackingFileSystem>,
-        conjure_http::private::Error,
-    > {
+            std::collections::BTreeMap<
+                String,
+                super::super::product::datasets::BackingFileSystem,
+            >,
+            conjure_http::private::Error,
+        > {
```
dtolnay, line 26: Fixed in prettyplease 0.1.11.

```suggestion
        std::collections::BTreeMap<
            String,
            super::super::product::datasets::BackingFileSystem,
        >,
        conjure_http::private::Error,
    > {
```

sfackler: Bumped, thanks!

PR #203, opened 2022-07-12
Make all types implement Ord, Eq, and Hash

### LINKED ISSUE #195: Objects with double keys can't be stored in sets
While we added `DoubleKey` to properly support `map<double, T>` and `set<double>`, (and aliases of doubles), we don't support `set<ObjectContainingADoubleField>` or `set<UnionContainingDoubleVariant>`.

The best option may be to make objects and unions unconditionally `Hash` + `Eq` + `Ord` by using manual impls.

sfackler:
## Before this PR
We had some special support for `double` values used as keys and set elements, but missed multiple corner cases.

## After this PR

==COMMIT_MSG==
All generated types implement `Ord`, `Eq`, and `Hash`. Notably, this enables support for `set<ObjectWithADoubleField>`.
==COMMIT_MSG==

We could have used the `DoubleKey` type everywhere and just used the normal `#[derive(Eq)]` etc, but I decided to make things a bit nicer for users by only using `DoubleKey` for `set<double>` and `map<double, T>` specifically. Aliases, objects, and unions wrapping `double` work with `f64` directly and have custom derive behavior.

## Possible downsides?
This is a breaking change. We could make it opt-in via a flag, but we're going to need to cut a major version bump anyways to pick up #200 so we might as well avoid the extra complexity.

Closes #195

### DIFF
--- a/conjure-object/Cargo.toml
+++ b/conjure-object/Cargo.toml
@@ -11,8 +11,15 @@ readme = "../README.md"
 [dependencies]
 base64 = "0.13"
 chrono = { version = "0.4", features = ["serde"] }
+educe = { version = "0.4", default-features = false, features = [
+    "Hash",
+    "PartialEq",
+    "Eq",
+    "PartialOrd",
+    "Ord",
+] }
 lazy_static = "1.0"
-ordered-float = { version = "3.0", features = ["serde"] }
+ordered-float = { version = "3", features = ["serde"] }
 regex = { version = "1.3", default-features = false, features = ["std"] }
 serde = "1.0"
 serde_bytes = "0.11"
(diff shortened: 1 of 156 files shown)


conjure-object/Cargo.toml
```diff
@@ -11,8 +11,15 @@ readme = "../README.md"
...
 chrono = { version = "0.4", features = ["serde"] }
+educe = { version = "0.4", default-features = false, features = [
+    "Hash",
+    "PartialEq",
+    "Eq",
+    "PartialOrd",
+    "Ord",
+] }
 lazy_static = "1.0"
-ordered-float = { version = "3.0", features = ["serde"] }
```
jonsyu1, line 15: Why did we drop the `.0` here?

jonsyu1 (approved):
Looks fine, just had a question about one of our dependency changes

conjure-object/Cargo.toml
sfackler: They're equivalent constraints - I think I just removed the dependency while writing the PR and then added it back later the other way.

PR #211, opened 2022-09-12
Support arbitrary response extensions

sfackler:
## Before this PR
We currently hardcode SafeParams as the only "special" bit of extra data that can be tracked regardless of the success or failure of an endpoint handler.

## After this PR

==COMMIT_MSG==
The `safe_params` argument to endpoint handlers has been replaced by `response_extensions` to allow for other custom response metadata.
==COMMIT_MSG==

In particular, we'll be able to use this for auditing in witchcraft-server.

### DIFF
--- a/conjure-codegen/src/servers.rs
+++ b/conjure-codegen/src/servers.rs
@@ -449,11 +449,12 @@ fn generate_endpoint_impl(
 
     let parts = quote!(parts_);
     let body = quote!(body_);
-    let safe_params = if has_safe_params(ctx, endpoint) {
-        quote!(safe_params_)
+    let response_extensions = if has_safe_params(ctx, endpoint) {
+        quote!(response_extensions_)
     } else {
-        quote!(_safe_params)
+        quote!(_response_extensions)
     };
+    let safe_params = quote!(safe_params_);
     let query_params = quote!(query_params_);
     let auth = quote!(auth_);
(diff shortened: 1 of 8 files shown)


conjure-codegen/src/servers.rs
```diff
@@ -449,11 +449,12 @@ fn generate_endpoint_impl(
 
     let parts = quote!(parts_);
     let body = quote!(body_);
-    let safe_params = if has_safe_params(ctx, endpoint) {
-        quote!(safe_params_)
+    let response_extensions = if has_safe_params(ctx, endpoint) {
+        quote!(response_extensions_)
```
dtobin, line 453: Is there a test that hits this code path? Looks like the generated test services don't have safe params.

conjure-http/src/server.rs
```diff
@@ -134,8 +135,8 @@ where
     #[allow(clippy::type_complexity)]
     fn handle<'life0, 'life1, 'async_trait>(
         &'life0 self,
-        safe_params: &'life1 mut SafeParams,
         req: Request<I>,
+        repsonse_extensions: &'life1 mut Extensions,
```
dtobin, line 139:
```suggestion
        response_extensions: &'life1 mut Extensions,
```

dtobin (approved):

PR #215, opened 2022-09-28
Support new log safety declarations

sfackler:
## Before this PR
We only supported detection of safe loggable endpoint parameters via the old `Safe` marker or `safe` tag.

## After this PR

==COMMIT_MSG==
Code generation now supports Conjure's new log safety declarations to track safe loggable endpoint parameters.
==COMMIT_MSG==

### DIFF
--- a/conjure-codegen/src/context.rs
+++ b/conjure-codegen/src/context.rs
@@ -1086,8 +1086,9 @@ impl Context {
                         .or_else(|| self.type_log_safety(f.type_()))
                 })
                 // The unknown variant is unsafe to log, and we don't want log safety to vary based
-                // on the type generation configuration.
-                .fold(Some(LogSafety::Unsafe), |a, b| self.combine_safety(a, b)),
+                // on the type generation configuration. However, like conjure-java we're going to
+                // treat the unknown variant as unannotated for now to ease the rollout.
+                .fold(None, |a, b| self.combine_safety(a, b)),
         };
 
         *ctx.log_safety.borrow_mut() = CachedLogSafety::Computed(safety.clone());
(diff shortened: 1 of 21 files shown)


conjure-codegen/src/context.rs
```diff
@@ -1002,6 +1027,84 @@ impl Context {
         }
     }
 
+    fn type_log_safety(&self, ty: &Type) -> Option<LogSafety> {
```
sfackler, line 1030: This logic should mirror what's defined in https://github.com/palantir/conjure-java/blob/develop/conjure-java-core/src/main/java/com/palantir/conjure/java/types/SafetyEvaluator.java.

conjure-codegen/src/context.rs
```diff
@@ -1002,6 +1028,84 @@ impl Context {
...
+                .union_()
+                .iter()
+                .map(|f| {
+                    f.safety()
+                        .cloned()
+                        .or_else(|| self.type_log_safety(f.type_()))
+                })
+                // The unknown variant is unsafe to log, and we don't want log safety to vary based
+                // on the type generation configuration.
+                .fold(Some(LogSafety::Unsafe), |a, b| self.combine_safety(a, b)),
```
dtobin, line 1090: Just acknowledging that this is a different decision than the Java implementation.

dtobin (approved):

conjure-codegen/src/context.rs
sfackler: Ah - I misread the Java logic! I will update to start with None.

dtobin: It was actually changed after the initial implementation: https://github.com/palantir/conjure-java/commit/4b08523c3b62455544f7b0544f07ae74eee4a13c

dtobin (approved):

PR #219, opened 2022-10-13
Support request context injection into server handlers

### LINKED ISSUE #216: Support opt-in request context injection
conjure-java supports the `server-request-context` tag at the endpoint level to add an extra `RequestContext` parameter to endpoint handlers. This can be used to add additional safe parameters to the request log entry, look at the raw header map, etc. We should support the same kind of thing.

sfackler:
## Before this PR
Conjure-generated handlers had no access to lower level information about the request, and in particular access to the request and response extensions for things like audit logging.

## After this PR

==COMMIT_MSG==
Added support for the `server-request-context` endpoint tag to provide lower level request information to the generated server handler.
==COMMIT_MSG==

Closes #216

### DIFF
--- a/changelog/@unreleased/pr-219.v2.yml
+++ b/changelog/@unreleased/pr-219.v2.yml
@@ -1,0 +1,6 @@
+type: feature
+feature:
+  description: Added support for the `server-request-context` endpoint tag to provide
+    lower level request information to the generated server handler.
+  links:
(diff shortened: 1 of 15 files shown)


dtobin (changes requested):
looks good other than one change

sfackler:
Fixed!

dtobin (approved):

PR #220, opened 2022-11-01
Add accessors to smile serializers and deserializers

sfackler:
Smile serialization and deserialization is stateful, so when in e.g. a nonblocking context you can need to update the inner IO type while preserving the state.

### DIFF
--- a/conjure-test/Cargo.toml
+++ b/conjure-test/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "conjure-test"
-version = "2.2.0"
+version = "3.2.0"
 authors = ["Steven Fackler <sfackler@palantir.com>"]
 edition = "2018"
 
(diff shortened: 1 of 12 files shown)


conjure-test/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "conjure-test"
-version = "2.1.0"
+version = "2.2.0"
```
dtobin, line 3: Should this be 3.2.0?

dtobin (approved):

conjure-test/Cargo.toml
sfackler: That crate isn't published so the version doesn't matter. It just ends up getting bumped when I use [cargo-edit](https://crates.io/crates/cargo-edit) to bump the crate versions. I should probably sync them up just to avoid confusion.

PR #224, opened 2022-12-02
Allow separate configuration of crate and endpoint versions

sfackler:
## Before this PR
The value used for the version of the generated crate and the service version in client endpoint metadata were always the same.

## After this PR

==COMMIT_MSG==
The value used for the version of the generated crate can now be set separately from the version used in client endpoint metadata.
==COMMIT_MSG==

We want the endpoint metadata to accurately reflect the version publishing the Conjure IR, but want the crate version to be static (e.g. `0.0.0`) to avoid unnecessarily modifying Cargo.lock.


TheKeveloper (approved):
Makes sense!

PR #227, opened 2022-12-13
Fix codegen for no-arg endpoints with a request context

sfackler:
## Before this PR
We would render syntactically invalid Rust for endpoints that had no arguments except for a request context. Rendering would bail out with an `expected expression` error.

## After this PR

==COMMIT_MSG==
Fixed code generation for endpoints with no arguments and a request context.
==COMMIT_MSG==


TheKeveloper (approved):
LGTM!

PR #241, opened 2023-04-27
Add a macro to make custom clients

sfackler:
## Before this PR
If you needed to talk to a non-Conjure API, you had to hand-write the client code.

## After this PR

==COMMIT_MSG==
Added a `#[conjure_client]` macro which will create client implementations for an annotated trait representing HTTP endpoints.
==COMMIT_MSG==

The annotation syntax is generally inspired by dialogue-annotations.


conjure-http/src/client.rs
```diff
@@ -223,3 +237,219 @@ pub trait AsyncWriteBody<W> {
...
+
+    /// Serializes the body.
+    fn serialize(value: T) -> Result<AsyncRequestBody<'a, W>, Error>;
+}
+
+/// A body serializer which encodes the type into JSON.
+///
+/// # Note
+///
+/// The default `serde_json` serializer is used, *not* the `conjure-serde` serializer.
```
sfackler, line 285: Does this seem like the right approach? We may want a `ConjureRequestSerializer` or something like that for things that talk to non-Conjure WC resources that uses our conjure serialization behavior?

conjure-http/src/client.rs
```diff
@@ -223,3 +237,219 @@ pub trait AsyncWriteBody<W> {
...
+pub enum JsonRequestSerializer {}
```
TheKeveloper, line 286: Maybe a dumb question, but why is this an enum vs a struct?

conjure-http/src/client.rs
TheKeveloper: nit: Maybe `SerdeRequestSerializer` to make it more explicit?

conjure-http/src/client.rs
sfackler: An empty enum can't be constructed, which we don't need to do in this context since none of the methods on the trait take `self`.

conjure-http/src/client.rs
sfackler: We definitely need to mention that it's JSON in the name. SerdeJsonRequestSerializer is more specific but maybe a bit wordy?

sfackler: I think I'm going to unify the standard Conjure codegen to use these same traits, so I think I'll actually just replace this with one that does the standard Conjure thing.

sfackler: Updated

conjure-macros/src/lib.rs
```diff
@@ -0,0 +1,805 @@
...
+    });
+
+    let Some(first) = it.next() else {
+        return Ok(Asyncness::Sync);
+    };
+
+    let is_async = first.sig.asyncness.is_some();
+
+    for f in it {
+        if f.sig.asyncness.is_some() != is_async {
```
TheKeveloper, line 275: Do we want to enforce that all of the methods in a custom client are either async or blocking? I'm not sure if there is a technical reason that we would need something like this, but I don't think that java dialogue client annotations enforce this?

TheKeveloper (approved):
I think this looks good and it seems like the tests pass! Just one question on whether we have to enforce that the entire trait is async instead of allowing individual methods to be async, but don't feel too strongly either way there.

This is super cool!

conjure-macros/src/lib.rs
sfackler: Unfortunately it has to be all or none - the generated code looks something like

```rust
pub struct MyServiceClient<T> {
    client: T,
}


impl<T> MyService for MyServiceClient<T>
where
    T: Client, // or T: AsyncClient
{
    ...
}
```

We need to bound the inner raw client type on either the async client trait or the blocking client trait, which means that all of the methods need to be the same asyncness.

In practice it shouldn't matter, since the asyncness of the generated client will depend on the asyncness of the server using it which is typically also an all-or-none thing.

sfackler: The Java logic handles this with the inner client always returning a `ListenableFuture` and the generated code either returns that or `.getUnchecked()` it, but asyncness is a more "viral" feature in Rust than our async Java clients.

Issue #245, opened 2023-05-01
Unify codegen for conjure/custom clients


sfackler:
We should have a shared code generation backend for clients generated from Conjure IR and custom ones generated from an annotated trait.

Issue #244, opened 2023-04-27
Support request writer, response body parameters in #[conjure_client]


sfackler:
You currently can't refer to the request body writer or response body types in a client trait definition, which makes working with streaming requests and responses pretty awkward. We should allow the traits to declare generic parameters to those types and hook them up properly in the impl.

PR #248, opened 2023-05-09
Add #[conjure_endpoints] for custom server endpoints

### LINKED ISSUE #231: Add macros for custom client/server interfaces
Some services need to interact with or expose APIs that cannot be defined in Conjure. Currently this requires a bunch of manual glue that we should be able to write for the service with a macro analagous to dialogue-annotations-processor:

```rust
#[service]
trait MyCustomService {
    #[endpoint(method = GET, path = "/foo/{bar}")]
    fn foo(&self, bar: i32) -> Result<(), Error>;
}
```

sfackler:
Same general approach as #241 but with some improvements that I'll port over to the client side in a follow up.

Closes #231

### DIFF
--- a/conjure-http/src/private/server.rs
+++ b/conjure-http/src/private/server.rs
@@ -49,7 +49,7 @@ where
     let params = value
         .split('/')
         .map(percent_encoding::percent_decode_str)
-        .map(Cow::from);
+        .map(|v| v.decode_utf8_lossy());
     D::decode(params).map_err(|e| e.with_safe_param("param", param))
 }
 
@@ -74,6 +74,18 @@ pub fn parse_query_params(parts: &request::Parts) -> HashMap<Cow<'_, str>, Vec<C
     }
 
     map
+}
+
+pub fn query_param<T, D>(
+    query_params: &HashMap<Cow<'_, str>, Vec<Cow<'_, str>>>,
+    key: &str,
+    param: &str,
+) -> Result<T, Error>
+where
+    D: DecodeParams<T>,
+{
+    let values = query_params.get(key).into_iter().flatten();
+    D::decode(values).map_err(|e| e.with_safe_param("param", param))
 }
 
 pub fn parse_query_param<T>(
--- a/conjure-http/src/private/server.rs
+++ b/conjure-http/src/private/server.rs
@@ -1,6 +1,6 @@
 use crate::private::{async_read_body, read_body, APPLICATION_JSON, APPLICATION_OCTET_STREAM};
 use crate::server::{
-    AsyncResponseBody, AsyncWriteBody, DecodeHeader, DecodeParams, ResponseBody, WriteBody,
+    AsyncResponseBody, AsyncWriteBody, DecodeHeader, DecodeParam, ResponseBody, WriteBody,
 };
 use crate::PathParams;
 use bytes::Bytes;
@@ -39,7 +39,7 @@ where
 
 pub fn path_param<T, D>(parts: &request::Parts, param: &str) -> Result<T, Error>
 where
-    D: DecodeParams<T>,
+    D: DecodeParam<T>,
 {
     let path_params = parts
         .extensions
@@ -82,7 +82,7 @@ pub fn query_param<T, D>(
     param: &str,
 ) -> Result<T, Error>
 where
-    D: DecodeParams<T>,
+    D: DecodeParam<T>,
 {
     let values = query_params.get(key).into_iter().flatten();
     D::decode(values).map_err(|e| e.with_safe_param("param", param))
--- a/conjure-http/src/private/server.rs
+++ b/conjure-http/src/private/server.rs
@@ -1,5 +1,7 @@
 use crate::private::{async_read_body, read_body, APPLICATION_JSON, APPLICATION_OCTET_STREAM};
-use crate::server::{AsyncResponseBody, AsyncWriteBody, ResponseBody, WriteBody};
+use crate::server::{
+    AsyncResponseBody, AsyncWriteBody, DecodeHeader, DecodeParams, ResponseBody, WriteBody,
+};
 use crate::PathParams;
 use bytes::Bytes;
 use conjure_error::{Error, InvalidArgument, PermissionDenied};
@@ -15,7 +17,7 @@ use std::borrow::Cow;
 use std::collections::{BTreeSet, HashMap};
 use std::error;
 
-const SERIALIZABLE_REQUEST_SIZE_LIMIT: usize = 50 * 1024 * 1024;
+pub const SERIALIZABLE_REQUEST_SIZE_LIMIT: usize = 50 * 1024 * 1024;
 
 pub fn parse_path_param<T>(parts: &request::Parts, param: &str) -> Result<T, Error>
 where
@@ -33,6 +35,22 @@ where
             Error::service_safe(e, InvalidArgument::new()).with_safe_param("param", param)
         })?;
     from_plain(&value, param)
+}
+
+pub fn path_param<T, D>(parts: &request::Parts, param: &str) -> Result<T, Error>
+where
+    D: DecodeParams<T>,
+{
+    let path_params = parts
+        .extensions
+        .get::<PathParams>()
+        .expect("PathParams missing from request");
+    let value = &path_params[param];
+    let params = value
+        .split('/')
+        .map(percent_encoding::percent_decode_str)
+        .map(Cow::from);
+    D::decode(params).map_err(|e| e.with_safe_param("param", param))
 }
 
 fn from_plain<T>(s: &str, param: &str) -> Result<T, Error>
@@ -154,6 +172,13 @@ where
     }
 
     Ok(())
+}
+
+pub fn header_param<T, D>(parts: &request::Parts, header: &str, param: &str) -> Result<T, Error>
+where
+    D: DecodeHeader<T>,
+{
+    D::decode(parts.headers.get_all(header)).map_err(|e| e.with_safe_param("param", param))
 }
 
 pub fn parse_required_header<T>(
--- a/conjure-http/src/server.rs
+++ b/conjure-http/src/server.rs
@@ -18,6 +18,7 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use conjure_error::{Error, InvalidArgument};
 use conjure_serde::json;
+use futures_core::Stream;
 use http::header::CONTENT_TYPE;
 use http::{
     request, Extensions, HeaderMap, HeaderValue, Method, Request, Response, StatusCode, Uri,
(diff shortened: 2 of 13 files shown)


conjure-http/src/server.rs
```diff
@@ -255,17 +269,30 @@ pub trait AsyncWriteBody<W> {
...
+    request_parts: MaybeBorrowed<'a, request::Parts>,
     response_extensions: &'a mut Extensions,
 }
 
 impl<'a> RequestContext<'a> {
     // This is public API but not exposed in docs since it should only be called by generated code.
     #[doc(hidden)]
     #[inline]
+    // FIXME remove in favor of borrowed constructor
     pub fn new(request_parts: request::Parts, response_extensions: &'a mut Extensions) -> Self {
```
sfackler, line 281: Some jank here but I don't want to make an internal-API breaking change since we upgrade conjure codegen and runtime libraries in separate excavators.

conjure-http/src/server.rs
```diff
@@ -300,3 +327,334 @@ impl<'a> RequestContext<'a> {
...
+    ///
+    /// The values have already been percent-decoded.
+    fn decode<I>(params: I) -> Result<T, Error>
+    where
+        I: IntoIterator,
+        I::Item: AsRef<str>;
+}
+
+/// A decoder which converts a single value using its [`FromStr`] implementation.
+pub enum FromStrDecoder {}
```
sfackler, line 516: On the client side we have separate types for header and params decoders, but that doesn't seem necessary so I've combined them here. If that seems right we can do some renames and deprecations in the client side.

conjure-macros/src/endpoints.rs
```diff
@@ -0,0 +1,705 @@
...
+    Type, Visibility,
+};
+
+pub fn generate(
+    _attr: proc_macro::TokenStream,
+    item: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let mut item = parse_macro_input!(item as ItemTrait);
+    let service = Service::new(&item);
+    strip_trait(&mut item);
```
sfackler, line 18: We now unconditionally strip helper attributes so that an error resolving the trait doesn't also cascade into a bunch of errors about unknown attributes. We'll want this in the client code as well.

conjure-macros/src/endpoints.rs
```diff
@@ -0,0 +1,705 @@
...
+
+    quote! {
+        <#serializer as conjure_http::server::#trait_<_, _>>::serialize(
+            &#parts.headers,
+            #response,
+        )
+    }
+}
+
+struct Service {
```
sfackler, line 468: This setup is moving towards a world in which we can use the same logic for custom and normal Conjure service generation re #245

conjure-macros/src/endpoints.rs
```diff
@@ -0,0 +1,705 @@
...
+    vis: Visibility,
+    name: Ident,
+    asyncness: Asyncness,
+    endpoints: Vec<Endpoint>,
+}
+
+impl Service {
+    fn new(trait_: &ItemTrait) -> Result<Self, Error> {
+        let mut errors = Errors::new();
```
sfackler, line 477: We do a better job of reporting all errors at once rather than bailing out on the first one compared to the client impl.

conjure-test/src/test/servers.rs
```diff
@@ -383,7 +402,7 @@ fn path_params() {
             Ok(())
         })
         .call()
-        .path_param("foo", "hello world")
+        .path_param("foo", "hello%20world")
```
sfackler, line 405: We should be testing that the params are percent-decoded by the decoder.

conjure-http/src/server.rs
```diff
@@ -300,3 +327,334 @@ impl<'a> RequestContext<'a> {
...
+pub trait DecodeHeader<T> {
+    /// Decodes the value from headers.
+    fn decode<'a, I>(headers: I) -> Result<T, Error>
+    where
+        I: IntoIterator<Item = &'a HeaderValue>;
+}
+
+/// A trait implemented by URL parameter decoders used by custom Conjure server trait
+/// implementations.
+pub trait DecodeParam<T> {
```
sfackler, line 505: This same trait is used for both path params and query params. Currently, path params will only ever see exactly one value in the iterator, but if we ever add support for multi-segment path parameters in the future this will be forwards compatible.

conjure-http/src/private/server.rs
```diff
@@ -35,6 +37,22 @@ where
...
+{
+    let path_params = parts
+        .extensions
+        .get::<PathParams>()
+        .expect("PathParams missing from request");
+    let value = &path_params[param];
+    let params = value
+        .split('/')
+        .map(percent_encoding::percent_decode_str)
+        .map(|v| v.decode_utf8_lossy());
```
sfackler, line 52: Not 100% sure what we want the right behavior to be here. In principle, you *could* allow path parameters to be arbitrary byte sequences after percent-decoding, but as far as I can tell no mainstream URI builders are built to support that.

Query parameters are actually specced to do a lossy UTF8 conversion, but as far as I can tell the same is not true for path parameters. We could instead do a non-lossy decode and fail the conversion which may be preferable?

conjure-macros/src/lib.rs
```diff
@@ -160,8 +162,239 @@ pub fn conjure_client(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///     async fn create_yak(
+///         &self,
+///         #[auth] auth: BearerToken,
+///         #[query(name = "parentName", decoder = FromStrOptionDecoder)] parent_id: Option<String>,
+///         #[body] yak: String,
+///     ) -> Result<(), Error>;
+/// }
+///
+/// #[conjure_endpoints]
+/// trait MyStreamingService<#[request_body] I, #[response_writer] O>
```
sfackler, line 281: This attribute approach can be ported into the client logic to handle #244

conjure-http/src/private/server.rs
```diff
@@ -35,6 +37,22 @@ where
...
 
+pub fn path_param<T, D>(parts: &request::Parts, param: &str) -> Result<T, Error>
+where
+    D: DecodeParam<T>,
+{
+    let path_params = parts
+        .extensions
+        .get::<PathParams>()
+        .expect("PathParams missing from request");
+    let value = &path_params[param];
```
TheKeveloper, line 48: This will panic if `param` is not in the map right? May want to modify `PathParams` to expose the `get` method of the underlying hash map and then map to an error. Unless for some reason `param` is guaranteed to be in here?

sfackler: Yeah, the param is guaranteed to be in there unless the actual server implementation is broken - all of the path params come out of the URI, and we shouldn't be able to route to the handler unless all of the params are present.

conjure-http/src/server.rs
```diff
@@ -300,3 +327,334 @@ impl<'a> RequestContext<'a> {
...
+        Self::check_content_type(headers)?;
+        let buf = private::async_read_body(body, Some(SERIALIZABLE_REQUEST_SIZE_LIMIT)).await?;
+        json::server_from_slice(&buf).map_err(|e| Error::service(e, InvalidArgument::new()))
+    }
+}
+
+/// A trait implemented by response serializers used by custom Conjure server trait implementations.
+pub trait SerializeResponse<T, W> {
+    /// Serializes the response.
+    fn serialize(request_headers: &HeaderMap, value: T)
```
TheKeveloper, line 412: I wonder if it may be worth passing in a struct that has the request headers and the values instead of passing them in as separate args? That way if there are other fields we care about in the future we can add those without making a breaking change to trait implementations

conjure-macros/src/lib.rs
```diff
@@ -160,8 +162,239 @@ pub fn conjure_client(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///         Defaults to `FromStrDecoder`.
+///     * `safe` - If set, the parameter will be added to the `SafeParams` response extension.
+/// * `#[query]` - A query parameter.
+///
+///     Parameters:
+///     * `name` - The string used as the key in the encoded URI. Required.
+///     * `decoder` - A type implementing `DecodeParam` which will be used to decode the value.
+///         Defaults to `FromStrDecoder`.
+///     * `safe` - If set, the parameter will be added to the `SafeParams` response extension.
+/// * `#[auth]` - A `BearerToken` used to authenticate the request.
```
TheKeveloper, line 202: Maybe worth mentioning that a client must send this as an `Authorization` header with value `Bearer <bearer-token>`

sfackler: It's mentioned down below in the parameter docs

conjure-http/src/server.rs
```diff
@@ -300,3 +327,334 @@ impl<'a> RequestContext<'a> {
...
+    fn serialize(
+        _request_headers: &HeaderMap,
+        value: T,
+    ) -> Result<Response<AsyncResponseBody<W>>, Error> {
+        Self::serialize_inner(value, AsyncResponseBody::Fixed)
+    }
+}
+
+/// A trait implemented by header decoders used by custom Conjure server trait implementations.
+pub trait DecodeHeader<T> {
```
TheKeveloper, line 496: For my own understanding, when should a sort of "static method" approach like this be preferred over something like,

```
pub trait FromHeader {
  fn decode(headers ...) -> Self
}
```
where the deserialization of a type is sort of "baked into" that type?

sfackler: That would mean there can only be one way to parse a header into some type, and mean that the only place any implementations for stdlib types could be defined is in conjure-http.

conjure-macros/src/lib.rs
TheKeveloper: Ah, missed that that was for the same field. Makes sense

conjure-http/src/server.rs
sfackler: I have this change partially implemented, but it's going to be a bit large. I think it probably makes sense to split out to a follow-up PR.

sfackler: https://github.com/palantir/conjure-rust/pull/258

conjure-http/src/server.rs
sfackler: For example, you may have an API that needs to hex-encode an integer rather than decimal-encoding it. Using separate decoder types allows the method itself to take a raw `i32` in both cases and just have differing annotations.

TheKeveloper: Ah, that makes sense! I was thinking that users would create newtypes in that case, but that would get pretty annoying

TheKeveloper (approved):

conjure-http/src/server.rs
sfackler: Yeah I figure it's better to add annotation burden rather than extra code at every call site.

sfackler:
:+1:

PR #250, opened 2023-05-10
Custom client improvements

### LINKED ISSUE #243: Clean up error reporting in #[conjure_client]
There are some situations where we bail out of the macro logic early due to an error, which then causes cascading errors all over the trait that can hide the real issue. Some more judicious error handling should be able to improve things.

### LINKED ISSUE #244: Support request writer, response body parameters in #[conjure_client]

sfackler:
This PR ports the improved logic used in #248 back over to the client macro.

Closes #243 
Closes #244


conjure-http/src/client.rs
```diff
@@ -328,7 +328,39 @@ pub trait AsyncDeserializeResponse<T, R> {
     fn accept() -> Option<HeaderValue>;
 
     /// Deserializes the response.
-    async fn deserialize(response: Response<R>) -> Result<T, Error>;
+    async fn deserialize(response: Response<R>) -> Result<T, Error>
+    where
+        R: 'async_trait;
```
sfackler, line 333: This isn't a breaking change, since implementations that don't need to support the shorter lifetime can omit the where clause.

TheKeveloper (approved):

Issue #182, opened 2022-03-16
Support dynamic configuration of conversion logic


sfackler:
All of the request and response conversion logic is currently hardcoded in `conjure_http::private::server`. In particular, this includes things like maximum request sizes and supported content types. We may want to make this configurable in a similar way to conjure-java's `UndertowRuntime`. This could probably be done backwards compatibly with a default type parameter?

sfackler:
The situation is similar on the client side.

PR #258, opened 2023-07-07
Add a runtime for server logic

sfackler:
While the `ConjureRuntime` is currently empty, it will enable us to support things like dynamic configuration of serializable encodings and custom request body limits.

cc #182

### DIFF
--- a/conjure-codegen/src/servers.rs
+++ b/conjure-codegen/src/servers.rs
@@ -288,7 +288,10 @@ fn generate_service_impl(ctx: &Context, def: &ServiceDefinition, style: Style) -
             T: #trait_name #params + 'static + #sync + #send,
             I: #input_trait<Item = #result<conjure_http::private::Bytes, conjure_http::private::Error>> #i_traits,
         {
-            fn endpoints(&self) -> #vec<#box_<dyn conjure_http::server::#endpoint_name<I, O> + Sync + Send>>
+            fn endpoints(
+                &self,
+                _: &conjure_http::private::Arc<conjure_http::server::ConjureRuntime>,
+            ) -> #vec<#box_<dyn conjure_http::server::#endpoint_name<I, O> + Sync + Send>>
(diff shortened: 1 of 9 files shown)


conjure-codegen/src/servers.rs
```diff
@@ -288,7 +288,10 @@ fn generate_service_impl(ctx: &Context, def: &ServiceDefinition, style: Style) -
             T: #trait_name #params + 'static + #sync + #send,
             I: #input_trait<Item = #result<conjure_http::private::Bytes, conjure_http::private::Error>> #i_traits,
         {
-            fn endpoints(&self) -> #vec<#box_<dyn conjure_http::server::#endpoint_name<I, O> + Sync + Send>>
+            fn endpoints(
+                &self,
+                _: &conjure_http::private::Arc<conjure_http::server::ConjureRuntime>,
```
sfackler, line 293: Conjure-generated server logic just ignores the runtime for now. When #245 is implemented it'll be able to take advantage of the runtime as well.

conjure-http/src/server.rs
```diff
@@ -468,61 +533,86 @@ impl ConjureResponseSerializer {
...
-    fn decode<I>(params: I) -> Result<T, Error>
+    fn decode<I>(&self, params: I) -> Result<T, Error>
     where
         I: IntoIterator,
         I::Item: AsRef<str>;
 }
 
 /// A decoder which converts a single value using its [`FromStr`] implementation.
-pub enum FromStrDecoder {}
+pub struct FromStrDecoder<'a> {
```
sfackler, line 599: Not sure if we'd ever want to pull dynamic logic for this kind of thing but I'm future proofing just in case.

sfackler:
I'll be doing the same thing for clients in a separate PR.

TheKeveloper (approved):
LGTM!

PR #264, opened 2023-09-20
fix: codegen succeeds for error type args that reference objects

afloren-palantir:
## Before this PR

When an error is define with an argument that references an object type the generator panics: https://app.circleci.com/pipelines/github/palantir/conjure-rust/1028/workflows/fc887808-5e16-4d64-8576-d457a7719813/jobs/2726

This is because on [this line](https://github.com/palantir/conjure-rust/pull/264/files#diff-e9427ea30d64f2b4aec4a2eac1240e285deec1c4d147da111020674b32333e22L332) we check the type context of `this_type` but it doesn't exist because `this_type` actually references the error. I think this is in part because we are re-using the object generation codepath with a made up "error type" [here](https://github.com/palantir/conjure-rust/blob/4df11248d7813a646e9bf1a6f027a91ef1280815/conjure-codegen/src/errors.rs#L27).

I also noticed while testing that the object arguments are not included in the serialized parameters of the generated error type. This appears to be intentional so leaving as is for now.

## After this PR
We inject the error types as virtual object definitions in the context object so that everything resolves correctly.

==COMMIT_MSG==
codegen succeeds for error type args that reference objects
==COMMIT_MSG==

## Possible downsides?

Technically you could reference an error type as in object in other fields/args but we expect this should be handled already when compiling from yml to the IR.

### DIFF
--- a/conjure-codegen/src/errors.rs
+++ b/conjure-codegen/src/errors.rs
@@ -18,18 +18,16 @@ use crate::context::Context;
 use crate::objects;
 use crate::types::{ErrorDefinition, ObjectDefinition};
 
-impl ErrorDefinition {
-    pub fn object_definition(&self) -> ObjectDefinition {
-        ObjectDefinition::builder()
-            .type_name(self.error_name().clone())
-            .fields(self.safe_args().iter().chain(self.unsafe_args()).cloned())
-            .docs(self.docs().cloned())
-            .build()
-    }
+pub fn error_object_definition(def: &ErrorDefinition) -> ObjectDefinition {
+    ObjectDefinition::builder()
+        .type_name(def.error_name().clone())
+        .fields(def.safe_args().iter().chain(def.unsafe_args()).cloned())
+        .docs(def.docs().cloned())
+        .build()
 }
 
 pub fn generate(ctx: &Context, def: &ErrorDefinition) -> TokenStream {
-    let object = def.object_definition();
+    let object = error_object_definition(def);
     let object_def = objects::generate(ctx, &object);
     let error_type = generate_error_type(ctx, def);
 
--- a/conjure-codegen/src/errors.rs
+++ b/conjure-codegen/src/errors.rs
@@ -18,18 +18,18 @@ use crate::context::Context;
 use crate::objects;
 use crate::types::{ErrorDefinition, ObjectDefinition};
 
-impl From<ErrorDefinition> for ObjectDefinition {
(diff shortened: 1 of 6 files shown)


conjure-codegen/src/errors.rs
```diff
@@ -18,12 +18,18 @@ use crate::context::Context;
...
 
+impl From<ErrorDefinition> for ObjectDefinition {
+    fn from(error: ErrorDefinition) -> Self {
+        ObjectDefinition::builder()
+            .type_name(error.error_name().clone())
+            .fields(error.safe_args().iter().chain(error.unsafe_args()).cloned())
+            .docs(error.docs().cloned())
+            .build()
+    }
+}
```
afloren-palantir, line 29: could be a regular helper fn if implementing `from` seems incorrect, I just wanted to ensure that the generated object definition was consistent between the code below and during context creation

sfackler: Yeah I think I'd just make this a function personally.

conjure-test/src/test/errors.rs
```diff
@@ -19,11 +19,17 @@ use crate::types::*;
 
 #[test]
 fn error_serialization() {
-    let error = SimpleError::new("hello", 15, false);
+    #[allow(deprecated)]
```
sfackler, line 22: What's deprecated?

afloren-palantir: the `TestObject` field `foo` is marked deprecated in the conjure but is also not optional, I'll just use a different test object

sfackler: Ah gotcha. Yeah probably best to use another object.

conjure-codegen/src/errors.rs
```diff
@@ -18,12 +18,18 @@ use crate::context::Context;
 use crate::objects;
 use crate::types::{ErrorDefinition, ObjectDefinition};
 
+impl ErrorDefinition {
+    pub fn object_definition(&self) -> ObjectDefinition {
```
sfackler, line 22: Can we instead make this a free function? I try to avoid adding new impls on generated types.

afloren-palantir: done, though lemme know if you have a preferred naming convention

sfackler (approved):

conjure-codegen/src/errors.rs
sfackler: LGTM!

PR #280, opened 2024-01-05
[stacked] Stop using #[async_trait] in conjure-http

sfackler:
## Before this PR
All async traits in the Conjure HTTP API had to be defined using the `#[async_trait]` macro. This caused the rustdoc output to be pretty unfriendly due to how the macro rewrote the trait definition, and required all futures to be boxed.

## After this PR

==COMMIT_MSG==
The async conjure-http traits no longer use the `#[async_trait]` macro.
==COMMIT_MSG==

With Rust 1.75's support for `impl trait` return types in traits, we can avoid the need for `#[async_trait]` and define the APIs in a more normal way.

Due to limitations in the current implementation in the compiler, we have to force all returned futures to be `Sync` in the trait definitions. This limitation will hopefully be lifted some time in 2024.

`impl trait`-returning methods are not object-safe, so the PR defines `BoxAsyncFoo` structs which type-erase an `AsyncFoo` implementation for the traits requiring that - in particular `AsyncEndpoint` and `AsyncWriteBody`.

The custom client and endpoints traits used with `conjure_macros` are still expected to use `#[async_trait]`. I will change that in a follow-up PR.

### DIFF
--- a/conjure-http/src/client.rs
+++ b/conjure-http/src/client.rs
@@ -272,14 +272,17 @@ impl<'a, W> BoxAsyncWriteBody<'a, W> {
     {
         BoxAsyncWriteBody { inner: Box::pin(v) }
     }
+}
 
-    /// Like [`AsyncWriteBody::write_body`].
-    pub async fn write_body(&mut self, w: Pin<&mut W>) -> Result<(), Error> {
+impl<W> AsyncWriteBody<W> for BoxAsyncWriteBody<'_, W>
+where
+    W: Send,
+{
+    async fn write_body(mut self: Pin<&mut Self>, w: Pin<&mut W>) -> Result<(), Error> {
         self.inner.as_mut().write_body(w).await
     }
 
-    /// Like [`AsyncWriteBody::reset`].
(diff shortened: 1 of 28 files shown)


conjure-http/src/client.rs
```diff
@@ -216,27 +209,82 @@ where
...
+    ) -> impl Future<Output = Result<(), Error>> + Send;
 
     /// Attempts to reset the body so that it can be written out again.
     ///
     /// Returns `true` if successful. Behavior is unspecified if this is not called after a call to `write_body`.
-    async fn reset(self: Pin<&mut Self>) -> bool
+    fn reset(self: Pin<&mut Self>) -> impl Future<Output = bool> + Send;
+}
+
+trait AsyncWriteBodyEraser<W> {
```
TheKeveloper, line 232: I know this isn't public, but can we add some docs for what this trait is supposed to be for? It's not very obvious to me

sfackler: Done!

TheKeveloper (approved):

PR #286, opened 2024-01-15
[stacked] Add conjure-specific deserializers and decoders

sfackler:
We will need these to port conjure-codegen over to using conjure-macros.

### DIFF
--- a/conjure-http/src/server/mod.rs
+++ b/conjure-http/src/server/mod.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 //! The Conjure HTTP server API.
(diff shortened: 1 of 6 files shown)


conjure-http/src/server/mod.rs
```diff
@@ -36,6 +35,10 @@ use std::str;
 use std::str::FromStr;
 use std::sync::Arc;
 
+use crate::private::{self, APPLICATION_JSON, SERIALIZABLE_REQUEST_SIZE_LIMIT};
+
+pub mod conjure;
```
sfackler, line 40: conjure-codegen will use types from both the top level module and the conjure submodule, but I figured it'd be best to split off the types that handle weird conjure-specific cases like `optional<binary>` and the `plain` encoding.

TheKeveloper (approved):

PR #290, opened 2024-01-20
[stacked] Allow configuration of the logged param name

sfackler:
We'll need this for conjure-codegen where we want to use the camelCase names of params in request logs and error reporting.


conjure-macros/src/lib.rs
```diff
@@ -271,13 +271,17 @@ pub fn conjure_client(attr: TokenStream, item: TokenStream) -> TokenStream {
 ///     * `decoder` - A type implementing `DecodeParam` which will be used to decode the value.
 ///         Defaults to `FromStrDecoder`.
 ///     * `safe` - If set, the parameter will be added to the `SafeParams` response extension.
+///     * `log_as` - The name of the parameter used in request logging and error reporting. Defaults
```
sfackler, line 274: I don't love this name but can't think of a better one. I don't expect anything will use this other than conjure-codegen anyway.

bavardage (dismissed):

TheKeveloper (approved):

PR #292, opened 2024-01-21
Convert endpoint codegen over to conjure-macros

sfackler:
The only user-visible change here should be that there are now separate `MyServiceEndpoints` and `AsyncMyServiceEndpoints` structs rather than just a single `MyServiceEndpoints` that handled both cases.

### DIFF
--- a/conjure-codegen/src/example_types/another/test_service.rs
+++ b/conjure-codegen/src/example_types/another/test_service.rs
@@ -1173,8 +1173,10 @@ where
         conjure_http::private::decode_empty_response(response_)
     }
 }
+use conjure_http::endpoint;
 ///A Markdown description of the service.
-pub trait TestService<I, O> {
+#[conjure_http::conjure_endpoints(name = "TestService")]
+pub trait TestService<#[request_body] I, #[response_writer] O> {
     ///The body type returned by the `get_raw_data` method.
     type GetRawDataBody: conjure_http::server::WriteBody<O> + 'static;
     ///The body type returned by the `get_aliased_raw_data` method.
@@ -1182,8 +1184,15 @@ pub trait TestService<I, O> {
     ///The body type returned by the `maybe_get_raw_data` method.
     type MaybeGetRawDataBody: conjure_http::server::WriteBody<O> + 'static;
     ///Returns a mapping from file system id to backing file system configuration.
+    #[endpoint(
+        method = GET,
+        path = "/catalog/fileSystems",
+        name = "getFileSystems",
+        produces = conjure_http::server::conjure::CollectionResponseSerializer
+    )]
     fn get_file_systems(
         &self,
+        #[auth]
         auth_: conjure_object::BearerToken,
     ) -> Result<
         std::collections::BTreeMap<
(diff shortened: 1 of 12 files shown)


conjure-codegen/src/example_types/another/test_service.rs
```diff
@@ -1173,17 +1173,26 @@ where
...
     type GetRawDataBody: conjure_http::server::WriteBody<O> + 'static;
     ///The body type returned by the `get_aliased_raw_data` method.
     type GetAliasedRawDataBody: conjure_http::server::WriteBody<O> + 'static;
     ///The body type returned by the `maybe_get_raw_data` method.
     type MaybeGetRawDataBody: conjure_http::server::WriteBody<O> + 'static;
     ///Returns a mapping from file system id to backing file system configuration.
+    #[endpoint(
+        method = GET,
+        path = "/catalog/fileSystems",
+        name = "getFileSystems",
```
pierceforte, line 1190: Why is it necessary to include a camelCase `name` here?

conjure-codegen/src/example_types/another/test_service.rs
```diff
@@ -1173,17 +1173,26 @@ where
...
+        produces = conjure_http::server::conjure::CollectionResponseSerializer
+    )]
     fn get_file_systems(
         &self,
+        #[auth]
         auth_: conjure_object::BearerToken,
```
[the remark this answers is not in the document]
sfackler: It avoids colliding with an argument that's also called `auth`. Conjure-defined params can never have trailing underscores.

conjure-codegen/src/example_types/another/test_service.rs
sfackler: Yeah - it ensures that the name included in trace/request logs matches what's in the Conjure definition.

conjure-codegen/src/example_types/another/test_service.rs
```diff
@@ -1192,2212 +1201,746 @@ pub trait TestService<I, O> {
...
+        #[path(
+            name = "datasetRid",
+            decoder = conjure_http::server::conjure::FromPlainDecoder,
+            log_as = "datasetRid"
+        )]
         dataset_rid: conjure_object::ResourceIdentifier,
     ) -> Result<std::collections::BTreeSet<String>, conjure_http::private::Error>;
+    #[endpoint(
+        method = GET,
+        path = "/catalog/datasets/{datasetRid}/branches/{branch:.+}/resolve",
```
pierceforte, line 1372: What does the `{branch:.+}` do?

conjure-codegen/src/example_types/another/test_service.rs
```diff
@@ -1192,2212 +1201,746 @@ pub trait TestService<I, O> {
...
         auth_: conjure_object::BearerToken,
+        #[body(deserializer = conjure_http::server::StdRequestDeserializer)]
         query: String,
+        #[query(
+            name = "different",
+            decoder = conjure_http::server::conjure::FromPlainDecoder
+        )]
         something: conjure_object::ResourceIdentifier,
+        #[query(
+            name = "optionalMiddle",
```
pierceforte, line 1467: I assume not, but is it maybe possible to specify the rename once for the function instead of each param?

conjure-codegen/src/example_types/another/test_service.rs
sfackler: It allows the param to match multiple path segments (e.g. `/branches/foo/bar/resolve`, `branch` would be `foo/bar`). It's not really used in modern conjure, and IIRC most implementations won't support it but I haven't updated the example definitions in a long time.

pierceforte (approved):
The example test service looks good to me. Just have some questions for my own understanding

conjure-codegen/src/example_types/another/test_service.rs
sfackler: yeah it has to be done on each param

pforte12 (approved):

PR #346, opened 2024-04-21
Remove fallible builders

### LINKED ISSUE #282: Add strict-staged builders
We currently support panicking and staged builders. We should remove support for the panicking ones since they're a bit dangerous, and add support for "strict staged" builders a la conjure-java where we don't special case collections.

This should allow us to remove all of the codegen in this crate in favor of the `staged-builder` crate.

sfackler:
## Before this PR
Consumers had to opt-into staged builders, and would otherwise get builders which would panic if required fields were not set.

## After this PR

==COMMIT_MSG==
Removed support for fallible builders.
==COMMIT_MSG==

Closes #282


TheKeveloper (approved):

sfackler:
@TheKeveloper build's green if you can reapprove.

TheKeveloper (approved):

PR #349, opened 2024-04-23
Use staged_builder

### LINKED ISSUE #347: Port builders to the staged_builder crate
https://docs.rs/staged-builder/latest/staged_builder/

### LINKED ISSUE #348: Limit constructors to only required fields
We currently create constructors for types which have 3 or fewer fields. The constructor treats all fields as required. This means that adding an optional field is a compile break: it will either add a new argument to the constructor or delete it.

Instead, we should ignore all non-required fields for the purpose of the constructor.

sfackler:
## Before this PR
We directly codegened staged builder logic, with each stage being a separate type.

Constructors were generated for all types which had 3 or fewer fields, and the constructor required all of them.

## After this PR

==COMMIT_MSG==
Object builders are now generated with `#[staged_builder]`. Constructors now ignore optional fields.
==COMMIT_MSG==

The builders generated by `#[staged_builder]` are much cleaner in rustdoc output, as all of the setters are rendered on one page: https://docs.rs/staged-builder/latest/staged_builder/example_person/struct.Builder.html

## Possible downsides?
The constructor change will be a someone nontrivial source break, but should make wire-compatible evolution of Conjure objects less likely to be a Rust-incompatible change.

Closes #347 
Closes #348


conjure-codegen/src/example_types/product/covariant_optional_example.rs
```diff
@@ -2,68 +2,23 @@ use conjure_object::serde::{ser, de};
 use conjure_object::serde::ser::SerializeStruct as SerializeStruct_;
 use std::fmt;
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[conjure_object::private::staged_builder::staged_builder]
+#[builder(crate = conjure_object::private::staged_builder, update, inline)]
 pub struct CovariantOptionalExample {
+    #[builder(default, into)]
     item: Option<conjure_object::Any>,
```
pierceforte, line 9: Why can it just do `into` here, whereas the Any type in a map or vec needs a customer serializer like https://github.com/palantir/conjure-rust/pull/349/files#diff-8fcc3688b43688b6b278cb5ba8f57cc9c0446f2b1200d948142255818816227cR11-R17?

pierceforte: As an aside, could we just define the custom serializer once somewhere so it doesn't have to keep redefining it?

conjure-codegen/src/example_types/product/enum_field_example.rs
```diff
@@ -2,67 +2,23 @@ use conjure_object::serde::{ser, de};
 use conjure_object::serde::ser::SerializeStruct as SerializeStruct_;
 use std::fmt;
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[conjure_object::private::staged_builder::staged_builder]
+#[builder(crate = conjure_object::private::staged_builder, update, inline)]
 pub struct EnumFieldExample {
+    #[builder()]
     enum_: super::EnumExample,
```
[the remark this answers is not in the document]
pierceforte: Sorry, I missed that this is how it's done on all fields.

conjure-codegen/src/example_types/product/covariant_optional_example.rs
pierceforte: I actually think this is specific to the `Option`, because `conjure_object::Any` has the custom serializer here https://github.com/palantir/conjure-rust/pull/349/files#diff-450a3ab87b025c7854669c7345e9221165642460bc03581346e2ed08db393726R11-R18

conjure-codegen/src/example_types/product/safe_long_example.rs
```diff
@@ -2,74 +2,23 @@ use conjure_object::serde::{ser, de};
 use conjure_object::serde::ser::SerializeStruct as SerializeStruct_;
 use std::fmt;
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Copy)]
+#[conjure_object::private::staged_builder::staged_builder]
+#[builder(crate = conjure_object::private::staged_builder, update, inline)]
 pub struct SafeLongExample {
+    #[builder()]
```
pierceforte, line 8: I notice that most of the `conjure_object` types don't need to specify `into` here. What makes that possible, whereas the `String` type does need the `into`?

conjure-codegen/src/example_types/product/covariant_optional_example.rs
sfackler: We can use `into` for places where we want the conversion logic to just use an `Into` implementation.

You can see in the deleted manual builder code below that the behavior here matches what we were doing previously.

conjure-codegen/src/example_types/product/safe_long_example.rs
sfackler: It depends on what conversion behavior we want to support for the setter. With an e.g. `i32` we just take that type directly, whereas for string we take `impl Into<String>` so you can pass in e.g. string literals directly.

sfackler: The empty `#[builder()]` attributes aren't necessary - I pushed up a commit to remove them.

pierceforte (approved):

pforte12 (approved):

PR #361, opened 2024-06-12
Add custom backtrace for conjure error

wsuppiger:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/conjure-error/src/error.rs
+++ b/conjure-error/src/error.rs
@@ -323,6 +323,13 @@ impl Error {
         self
     }
 
+    /// Adds a new custom backtrace to the error which is safe to log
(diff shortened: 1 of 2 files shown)


conjure-error/src/error.rs
```diff
@@ -392,17 +399,30 @@ impl<'a> Iterator for ParamsIter<'a> {
 
 /// A backtrace associated with an `Error`.
 // FIXME remove in favor of std backtrace directly in next major bump
```
sfackler, line 401: We can remove this FIXME now that our type also supports custom traces.

sfackler (approved):

PR #380, opened 2024-09-16
Support per-endpoint request size limits

### LINKED ISSUE #374: Support per-endpoint request size limits
We currently apply a global 50MiB limit on serializable request bodies in the server logic. We should support per-endpoint size limits like conjure-java does. Unfortunately, this may require an API break to the request deserializer APIs :(

sfackler:
## Before this PR
All endpoints with serializable request bodies limited them to 50MiB without the ability to configure that.

## After this PR

==COMMIT_MSG==
Added support for per-endpoint request size limits via the `server-limit-request-size: N` tag.
==COMMIT_MSG==

There's a bit of jank in the codegen since up until now it's been totally infallible. To work around this, we inject `compile_error!` macro calls when the configured limit fails to parse properly. In the Java implementation, this parsing is deferred to runtime.

The limit is passed as a const generic parameter in the `StdRequestDeserializer`. I'm not sure if this is the "right" approach, but it lets us preserve back compat. Custom endpoints can set the parameter directly, which is workable but a bit weird. In the future it might make sense for the limit to be a more first-class concept in some way.

Closes #374

### DIFF
--- a/conjure-codegen/src/human_size.rs
+++ b/conjure-codegen/src/human_size.rs
@@ -1,0 +1,50 @@
+pub fn parse(s: &str) -> Result<usize, String> {
+    let split = s.find(|c: char| !c.is_ascii_digit()).unwrap_or(s.len());
+    let (number, unit) = s.split_at(split);
+
+    let number = number.trim().parse::<usize>().map_err(|e| e.to_string())?;
+
+    let multiple = match &*unit.trim().to_ascii_lowercase() {
+        "b" | "" => 1,
+        "k" | "kb" => 1000,
+        "ki" | "kib" => 1024,
+        "m" | "mb" => 1000 * 1000,
+        "mi" | "mib" => 1024 * 1024,
+        "g" | "gb" => 1000 * 1000 * 1000,
+        "gi" | "gib" => 1024 * 1024 * 1024,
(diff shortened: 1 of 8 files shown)


conjure-codegen/src/human_size.rs
```diff
@@ -0,0 +1,50 @@
+pub fn parse(s: &str) -> Result<usize, String> {
```
TheKeveloper, line 1: Is this based on an equivalent java class where we are trying to maintain the same logic? Or is there some crate we could use that handles this so we're not worried about edge cases?

The logic here seems fine to be but I'm a bit paranoid that there's some edge case we've missed.

sfackler: Yeah, the Java version is here: https://github.com/palantir/human-readable-types/blob/develop/human-readable-types/src/main/java/com/palantir/humanreadabletypes/HumanReadableByteCount.java. It should be the same except the Rust version doesn't currently handle `byte`, `mebibyte`, etc since I never see those used in practice.

sfackler: I haven't found an existing crate that handles this parsing unfortunately.

TheKeveloper (approved):

PR #407, opened 2024-11-22
Begin implementing OpenAPI -> Rust codegen

jgrazian:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Cargo.lock
 .idea/
 *.iml
 
+/conjure-openapi/out/**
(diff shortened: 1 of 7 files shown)


sfackler:
This seems like a reasonable thing to build but I don't think it should be done in this repo in particular.

jgrazian:
@sfackler Thanks for the comment, this is fine. I was having trouble trying to figure out how these changes would play nice with the RFC-002 anyway.

Are you open to the changes exposing the `conjure-codegen` types and `generate_files_inner` so that this can be moved externally?

sfackler:
I definitely don't want to make conjure-codegen internals public API.

Issue #283, opened 2024-01-08
Add a client ConjureRuntime


sfackler:
Same idea as the server side one.

PR #410, opened 2024-12-29
Support runtime configuration of server encodings

### LINKED ISSUE #182: Support dynamic configuration of conversion logic

### LINKED ISSUE #249: Support SMILE in serializable responses
conjure-serde has a SMILE encoder, but we currently don't use it in conjure-http.

sfackler:
## Before this PR
The server only supported JSON-encoded serializable request and response bodies.

## After this PR

==COMMIT_MSG==
The server now supports SMILE-encoded serializable request and response bodies by default. The set of supported encodings can be configured in `ConjureRuntime`.
==COMMIT_MSG==

Rust clients still only request JSON - fixing that depends on #283.

## Possible downsides?
We previously did no validation of the `Accept` header when serializing the response body. Clients that are providing `Accept` headers which don't actually declare acceptance of JSON or SMILE will break.

We're now using a type-erased wrapper around serde for serialization and deserialization, which could result in slower performance.

Closes #182
Closes #249


conjure-http/src/server/runtime.rs
```diff
@@ -0,0 +1,391 @@
...
+    /// matched against the same MIME type, the encoding registered first by [`Builder::encoding`] will be selected.
+    ///
+    /// The implementation treats the absense of the `Accept` header equivalently to `Accept: */*`.
+    ///
+    /// [RFC 9110]: https://httpwg.org/specs/rfc9110.html#field.accept
+    pub fn response_body_encoding(
+        &self,
+        headers: &HeaderMap,
+    ) -> Result<&(dyn Encoding + Sync + Send), Error> {
+        let mut types = headers
```
sfackler, line 93: conjure-java has much simpler logic for this but I figured we may as well support the RFC-mandated behavior more closely. We still ignore parameters though.

conjure-test/src/test/servers.rs
```diff
@@ -595,6 +595,19 @@ fn optional_alias_request() {
...
 
+#[test]
+fn smile_request() {
+    TestServiceHandler::new()
+        .json_request(|body| {
+            assert_eq!(body, "hello world");
+            Ok(())
+        })
+        .call()
+        .body(b":)\n\x05Jhello world")
```
bavardage, line 606: heh I didn't know smile uses `:)` as header bytes

conjure-http/src/server/runtime.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    /// Returns the appropriate [`Encoding`] to serialize the response body.
+    ///
+    /// The MIME types in the request's `Accept` header are processed in accordance with [RFC 9110]. If two MIME types
+    /// have equal preference by that algorithm, the implementation will prefer the type declared earlier in the header.
+    ///
+    /// MIME types are matched against [`Encoding::content_type`], ignoring parameters. If multiple [`Encoding`]s are
+    /// matched against the same MIME type, the encoding registered first by [`Builder::encoding`] will be selected.
+    ///
+    /// The implementation treats the absense of the `Accept` header equivalently to `Accept: */*`.
```
bavardage, line 86: nit: s/absense/absence/

conjure-http/src/server/runtime.rs
```diff
@@ -0,0 +1,391 @@
...
+
+/// A type providing server logic that is configured at runtime.
+pub struct ConjureRuntime {
+    encodings: Vec<Box<dyn Encoding + Sync + Send>>,
+}
+
+impl ConjureRuntime {
+    /// Creates a new runtime with default settings.
+    pub fn new() -> Self {
+        Self::builder().build()
```
bavardage, line 37: ok cool, in the builder adds json and smile if no encodings added explicitly

sfackler: Yeah the only behavior change unless you start messing with config is that the server will be willing to consume Smile requests and return Smile responses if asked, but it'll still default to JSON.

bavardage (approved):

bavardage (approved):

PR #412, opened 2025-01-03
Derive Serialize and Deserialize for most objects

sfackler:
## Before this PR
We manually implemented Serialize and Deserialize for all types, even if their representations could be handled by the normal serde derive macros.

## After this PR

==COMMIT_MSG==
Serialize and Deserialize implementations of objects, aliases, and enums now use serde's derive macros.
==COMMIT_MSG==

This both shrinks the amount of code we're generating and should make it a bit more readable.

Unions are still manually implemented since serde doesn't support Conjure's weird combination of external and adjacent tagging. At some point we should move that logic into our own `#[derive(Serialize_conjure)]` macro.


conjure-codegen/src/example_types/product/alias_as_map_key_example.rs
```diff
@@ -18,6 +27,11 @@ pub struct AliasAsMapKeyExample {
         default,
         map(key(type = super::RidAliasExample), value(type = super::ManyFieldExample))
     )]
+    #[serde(
+        rename = "rids",
+        skip_serializing_if = "std :: collections :: BTreeMap :: is_empty",
```
bavardage, line 32: are the spaces here ok? intentional?

bavardage: (not mentioned in https://serde.rs/field-attrs.html)

sfackler: It works, but is just a big ugly. I'll fix.

bavardage: yeah guess `quote!` just puts them in and they don't hurt

sfackler: Fixed!

bavardage (approved):

PR #415, opened 2025-01-20
Split out separate top level modules

### LINKED ISSUE #279: Partition objects/errors/clients/endpoints into separate top-level modules
We currently put the generated code for all Conjure types together in the same module. However, once we implement #121, we'll end up generating two traits for each endpoint, and there isn't a clear naming convention I can think of to disambiguate.

We should update the codegen to create separate top-level modules for each Conjure type to keep them separate from each other. For example, using the example API, we would have

```rust
use example_api::another::{TestServiceClient, TestService, TestServiceEndpoints, DifferentPackage};
```
change to
```rust
use example_api::{
    clients::another::TestServiceClient,
    endpoints::another::{TestService, TestServiceEndpoints},
    objects::another::DifferentPackage,
};
```

We could go further and split them into separate `example_api_objects`, `example_api_clients`, etc crates like conjure-java does, but that doesn't really seem worth it IMO.

sfackler:
## Before this PR
Generated objects, errors, clients, and endpoints were all included together in the same module hierarchy. This opens up the possibility of name collisions (e.g. if you have a service called `Foo` and an object called `FooEndpoints`), and is a blocker for conversion of client code generation to use the new macro.

## After this PR

==COMMIT_MSG==
Generated code is now split into separate top-level modules for objects, errors, clients, and endpoints.
==COMMIT_MSG==

Closes #279


conjure-codegen/src/lib.rs
```diff
@@ -279,13 +279,17 @@
 //!
 //! * `server-request-context` - The generated server trait method will have an additional
 //!     `RequestContext` argument providing lower level access to request and response information.
+//! * `server-limit-request-size: <size>` - Sets the maximum request body size for endpoints with
+//!     serializable request bodies. `<size>` should be a human-readable byte count (e.g. `50Mi` or
+//!     `100Ki`). Defaults to `50Mi`.
```
sfackler, line 284: I forgot to document this in #380

ellisjoe (approved):

ellisjoe (approved):

PR #417, opened 2025-01-21
[stacked 2/n] Rewrite client codegen to use #[conjure_client]

sfackler:
## Before this PR
Conjure-generated clients were still open-coded rather than using the new `#[conjure_client]` macro.

## After this PR

==COMMIT_MSG==
Conjure-generated clients now use the `#[conjure_client]` macro. Consumers will need to import the new `FooService` trait to use the `FooServiceClient` methods. The async client has been renamed from `FooServiceAsyncClient` to `AsyncFooServiceClient`.
==COMMIT_MSG==


conjure-test/src/test/clients.rs
```diff
@@ -453,7 +453,6 @@ fn unexpected_json_response() {
 fn json_request() {
     let client = TestClient::new(Method::POST, "/test/jsonRequest")
         .header("Content-Type", "application/json")
-        .header("Content-Length", "13")
```
ellisjoe, line 456: Why is this no longer necessary?

ellisjoe (approved):

ellisjoe (approved):

Issue #418, opened 2025-01-21
Serializer error parameters properly


sfackler:
We currently only support serialization of scalar error parameters and stringify them to align with a longstanding bug in the conjure-java. This is now being resolved - errors registered explicitly with endpoints are now serialized properly.

We should figure out how to make this change as well. Like in the Java implementation, we'll have to be careful to not break backwards compatibility with clients expecting to receive stringified parameters.

PR #420, opened 2025-01-22
Overhaul the ErrorType trait

### LINKED ISSUE #339: Provide static access to error names
Unlike conjure-rust, it's currently hard to tell if a `SerializableError` is a specific Conjure error type. The error name of a Conjure error is only accessible from an instance of that type, which we don't have.

sfackler:
## Before this PR
The `ErrorType` trait had some significant limitations (like the inability to statically retrieve an error type's name string), and retained functionality (like with_instance_id) that hasn't been used in many years.

## After this PR

==COMMIT_MSG==
Overhauled the `ErrorType` trait. Its methods are now static, allowing easier comparison of the type of a serialized error.
==COMMIT_MSG==

When we work out #418, we can add some utility methods to turn a `SerializableError` back into a typed error.

Closes #339

### DIFF
--- a/conjure-error/src/lib.rs
+++ b/conjure-error/src/lib.rs
@@ -21,7 +21,7 @@
 extern crate self as conjure_error;
 
 use conjure_object::Uuid;
-use serde::{Serialize, Serializer};
+use serde::Serialize;
 
 use crate::ser::{ParametersSerializer, StringSeed};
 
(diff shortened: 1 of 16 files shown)


conjure-error/src/lib.rs
```diff
@@ -57,102 +57,40 @@ impl ErrorCode {
...
     /// Returns the error's name.
     ///
     /// The name must be formatted like `NamespaceName:ErrorName`.
-    fn name(&self) -> &str;
-
-    /// Returns the error's instance ID, if it stores one.
-    ///
-    /// Conjure-generated error types return `None`, but other implementations like those for `SerializableError`
-    /// and `WithInstanceId` return a value.
-    fn instance_id(&self) -> Option<Uuid>;
```
sfackler, line 71: We do need the ability to manually set the instance ID for https://github.com/palantir/conjure-rust-runtime/pull/235, but this old method still doesn't feel like the right place to put it...

PR #422, opened 2025-01-22
[stacked 4/n] Clean up deprecations and FIXMEs

sfackler:


conjure-object/src/private.rs
```diff
@@ -216,7 +216,7 @@ where
 }
 
 // FIXME make private in 5.0
```
amazr, line 218: Probably want to remove this comment?

amazr (approved):

PR #442, opened 2025-05-04
Allow non-Send clients

sfackler:
## Before this PR
The structs generated by the `#[conjure_client]` macro always required the inner client type to be `Send`. However, clients in wasm targets backed by the Javascript `fetch` API will not be `Send` and can't be cleanly used.

## After this PR

==COMMIT_MSG==
Added a set of non-Send async client APIs.
==COMMIT_MSG==

I'm not updating the conjure-codegen logic now, since that'll be backed by the macro once #417 lands.

## Possible downsides?
This adds a third variant of the client APIs (on top of the old blocking and Send-requiring async versions) that are almost but not perfectly identical. In the future, the mandatory Send injection can probably removed once return type notation is stabilized (e.g. `where <T as AsyncClient>::send(): Send`).

### DIFF
--- a/conjure-http/src/client.rs
+++ b/conjure-http/src/client.rs
@@ -122,6 +122,16 @@ pub enum AsyncRequestBody<'a, W> {
     Streaming(BoxAsyncWriteBody<'a, W>),
 }
 
+/// The body of a local async Conjure request.
+pub enum LocalAsyncRequestBody<'a, W> {
+    /// No body.
+    Empty,
+    /// A body already buffered in memory.
+    Fixed(Bytes),
+    /// A streaming body.
+    Streaming(BoxLocalAsyncWriteBody<'a, W>),
+}
+
 /// A trait implemented by HTTP client implementations.
 pub trait Client {
     /// The client's binary request write type.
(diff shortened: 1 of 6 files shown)


conjure-http/src/client.rs
```diff
@@ -444,6 +622,24 @@ where
...
+    R: Stream<Item = Result<Bytes, Error>>,
+{
+    fn accept() -> Option<HeaderValue> {
+        Some(APPLICATION_JSON)
+    }
+
+    async fn deserialize(response: Response<R>) -> Result<T, Error> {
+        if response.headers().get(CONTENT_TYPE) != Some(&APPLICATION_JSON) {
+            return Err(Error::internal_safe("invalid response Content-Type"));
+        }
```
carterkozak, line 637: This logic matches existing code in `AsyncDeserializeResponse` and `DeserializeResponse` as well as `decode_serializable_response` and `async_decode_serializable_response`, but I don't think it's quite right. It's common for servers to respond with something like `Content-Type: application/json; charset=utf-8` or similar (allowing additional arbitrary key/value pairs of parameters.

We take a shortcut in dialogue using prefix matching, which isn't a general solution, but allows us to support these cases without much complexity: https://github.com/palantir/dialogue/blob/5567efd8f6507c0218279d341cdbefb7d8043a7c/dialogue-annotations/src/main/java/com/palantir/dialogue/annotations/Json.java#L116-L121
This is justifiable because we support a relatively small set of encodings, none of which are prefixes of others that we're aware of. I do think it's generally preferable here to err on the side of assuming data is the correct shape, as deserialization will fail if the content doesn't match. The infra folks use in the wild is all over the place...

carterkozak (approved):

PR #456, opened 2025-06-04
Add '--extraManifestJson' arg

nlcamp:
## Before this PR

The only editable fields in the generated Cargo.toml were `name` and `version`. There was no way to specify other manifest fields such as `publish`.

## After this PR

==COMMIT_MSG==
Adds `--extraManifestJson` arg, which allows specifying a JSON blob that will be converted to TOML and added to the generated Cargo.toml.
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Options:
           The version of the generated crate. Defaults to `--productVersion`
       --extraManifestJson <json>
           Extra manifest configuration as a JSON object.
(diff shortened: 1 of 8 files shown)


JohnnyGOX17 (dismissed):

sfackler (dismissed):
Thanks!

sfackler (approved):
Thanks!

PR #460, opened 2025-06-27
Add a CLI flag to enable using external references

markelliot:
## Before this PR
conjure-rust does not support external references

## After this PR
==COMMIT_MSG==
Add `--use-external-references` to resolve external references to Rust types.
==COMMIT_MSG==

## Possible downsides?
Opt-in, should be very little risk. Conjure only supports `java` as an `external` type key but the IR contains `externalReference` as a universal type, so users may find the config to generate IR where this feature is usable in conjure-rust a little weird.

### DIFF
--- a/conjure-codegen/src/context.rs
+++ b/conjure-codegen/src/context.rs
@@ -42,6 +42,7 @@ pub struct Context {
     exhaustive: bool,
     serialize_empty_collections: bool,
     strip_prefix: Vec<String>,
+    use_external_references: bool,
     version: Option<String>,
 }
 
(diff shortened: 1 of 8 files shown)


sfackler:
Hey melliot!

I don't think we really want to expand the use of external references - they exist purely as a pretty horrible hack to get Java-based APIs moved over to more easily from the early days of Conjure.

Depending on your setup, there are some potential postprocessing hacks you could use. For example, you could define "fake" stand-in types in a specific module which you could then delete and replace with one that reexports the proper non-Conjure types.

PR #476, opened 2025-07-11
Expand on conjure-error docs

sfackler:
Previously, it could be pretty hard to figure out how errors in Rust conjure services were supposed to "work". This should make them a bit more understandible.

### DIFF
--- a/conjure-error/src/lib.rs
+++ b/conjure-error/src/lib.rs
@@ -14,8 +14,93 @@
 
 //! Runtime support for Conjure error types.
 //!
-//! Conjure errors are represented by a struct implementing the `ErrorType` trait. The struct's fields are the error's
-//! parameters, and the trait implementation stores the remainder of the error's information.
+//! In a networked service, the error objects that are propagated through its codebase are responsible for two things:
+//!
+//! * Collecting useful information that a developer can use to diagnose whatever problem caused the error.
+//! * Controlling how the error is presented to the client.
+//!
+//! Services implemented using Conjure's frameworks use the [`Error`] type defined in this crate as the single error
+//! type throughout the codebase. [`Error`]s store:
+//!
+//! * Developer facing:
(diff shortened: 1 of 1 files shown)


conjure-error/src/lib.rs
```diff
@@ -14,8 +14,93 @@
...
+//! Services implemented using Conjure's frameworks use the [`Error`] type defined in this crate as the single error
+//! type throughout the codebase. [`Error`]s store:
+//!
+//! * Developer facing:
+//!   * [`Error::cause`] - the underlying cause of the error. This can be a type implementing the Rust
+//!     [`std::error::Error`] trait, or just a [`str`] or [`String`] containing a description of what happened. When
+//!     an [`Error`] is logged, the cause (and its chain of sources via [`std::error::Error::source`]) are included as a
+//!     parameter. The log-safety of that cause information is identified by the choice of constructor of the [`Error`].
+//!   * [`Error::safe_params`] and [`Error::unsafe_params`] - key-value pairs that can be added to the error to provide
+//!     context. Then an [`Error`] is logged, these are included in the service log's parameters. When a service
```
arseny-bogomolov, line 31:
```suggestion
//!     context. When an [`Error`] is logged, these are included in the service log's parameters. When a service
```

conjure-error/src/lib.rs
```diff
@@ -14,8 +14,93 @@
...
+//!     [`Error`] is created, all of the parameters of its associated Conjure error are automatically included as
+//!     params. Additional params can be added via [`Error::with_safe_param`] and [`Error::with_unsafe_param`].
```
arseny-bogomolov, line 33: Can you be a bit clearer here: as I understand it, it's the `ErrorType::safe_args()` that make it into`Error::safe_params` by default, but to me at least that's not immediately clear from "all of the parameters of its associated Conjure error"

sfackler: Updated!

PR #477, opened 2025-07-24
improvement: special case the builder build keyword for object fields

overtonch:
## Before this PR

Generating a conjure object with a field named `build` succeeds, but the generated rust code will not compile since `build` conflicts with the builder's build method.

## After this PR

==COMMIT_MSG==
The `build` keyword is special-cased alongside other rust language keywords to be prefixed with an underscore on the generated builder object.
==COMMIT_MSG==

### DIFF
--- a/conjure-codegen/src/context.rs
+++ b/conjure-codegen/src/context.rs
@@ -886,6 +886,8 @@ impl Context {
             | "typeof" | "unsized" | "virtual" | "yield" => true,
             // weak keywords
             "union" | "dyn" => true,
+            // builder pattern methods
+            "build" => true,
             _ => false,
         };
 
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/conjure-rust](https://github.com/palantir/conjure-rust), @overtonch! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

conjure-codegen/src/context.rs
```diff
@@ -886,6 +886,8 @@ impl Context {
             | "typeof" | "unsized" | "virtual" | "yield" => true,
             // weak keywords
             "union" | "dyn" => true,
+            // builder pattern methods
+            "build" => true,
```
sfackler, line 890: We'll also want to cover `new` and `builder` I think to cover those static methods: https://docs.rs/conjure-codegen/latest/conjure_codegen/example_types/product/struct.SetExample.html

sfackler (approved):

PR #479, opened 2025-07-25
Add wildcard support for conjure-macros + fix gradle for publish

amazr:
## Before this PR

A couple of things:
- rust witchcraft supports routing to wildcards, and regex matches in general
- conjure codegen supports generating rust conjure with wildcard syntax
- the above codegen generates code that's broken at runtime, because conjure-rust does not plug into rust witchcrafts regex routing, and there is no path param decoder that preserves the full path segments
- gradle publish is broken on develop

## After this PR

==COMMIT_MSG==
Add wildcard support for conjure-macros and add wildcard path decoder + fix gradle for publish
==COMMIT_MSG==

### DIFF
--- a/conjure-http/src/server/mod.rs
+++ b/conjure-http/src/server/mod.rs
@@ -769,28 +769,6 @@ where
     }
 }
 
-/// A decoder which converts a single value using its [`FromStr`] implementation.
-pub enum FromWildcardDecoder {}
-
-impl<T> DecodeParam<T> for FromWildcardDecoder
-where
-    T: FromStr,
-    T::Err: Into<Box<dyn error::Error + Sync + Send>>,
-{
-    fn decode<I>(_: &ConjureRuntime, params: I) -> Result<T, Error>
-    where
-        I: IntoIterator,
-        I::Item: AsRef<str>,
-    {
-        let s = params
-            .into_iter()
-            .map(|p| p.as_ref().to_owned())
-            .collect::<Vec<_>>()
-            .join("/");
-        T::from_str(&s).map_err(|e| Error::service(e, InvalidArgument::new()))
-    }
-}
-
 fn optional_item<I>(it: I) -> Result<Option<I::Item>, Error>
 where
     I: IntoIterator,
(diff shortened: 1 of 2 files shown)


sfackler:
Yeah, we're going to need to tweak the codegen a bit - we need to separate out the name and regex components and forward them along appropriately here: https://github.com/palantir/conjure-rust/blob/master/conjure-macros/src/endpoints.rs#L229

conjure-test/src/test/servers.rs
```diff
@@ -949,6 +958,17 @@ fn custom_path_params() {
...
 
+#[test]
+fn wildcard_path_param() {
+    let mut mock = MockCustomService::new();
+    mock.expect_wildcard_path_param()
+        .with(eq("some/wildcard/match".to_string()))
+        .returning(|_| Ok(()));
+    Call::new(CustomServiceEndpoints::new(mock))
+        .path_param("path", "some/wildcard/match")
+        .send_sync("wildcard_path_param");
```
sfackler, line 969: Can you also have this look at the PathSegments and confirm that the regex pattern is set there?

amazr: Added that, also ended up realizing that you have a regex in one of the client macros in a test. I know we chatted about not allowing those for clients, curious if you have further thoughts on that?

amazr: Sorry, not for clients, it's here where we have a ".+" regex pattern: https://github.com/palantir/conjure-rust/blob/b900cffd553a1a9f6a82eacd109e9a21af849a78/conjure-codegen/src/example_types/another/test_service.rs#L1372

conjure-http/src/server/mod.rs
```diff
@@ -769,6 +769,28 @@ where
     }
 }
 
+/// A decoder which converts a single value using its [`FromStr`] implementation.
+pub enum FromWildcardDecoder {}
```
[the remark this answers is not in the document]
amazr: ah, didn't catch that, but that definitely works!

sfackler (approved):

sfackler:
Thanks! If you backport to the release/4.x.x branch I can cut a release.

PR #480, opened 2025-07-28
Expose conjure tags as endpoint parameters

atrocities:
Tags were previously not exposed at the endpoint level. The only tags that were effective were 'server-request-context' and 'server-limit-request-size', both of which were interpreted by conjure-rust at codegen time.

Add a 'tags' function to Endpoint, which exposes the set of tags. This is similar to the java undertow generator.

Forgoes adding 'tags' to the client, unlike the java dialogue implementation. This is because the clients are used directly in rust, as opposed to being passed into a dialogue client generator.

### DIFF
--- a/conjure-codegen/src/servers.rs
+++ b/conjure-codegen/src/servers.rs
@@ -6,6 +6,7 @@ use crate::types::{
 use heck::ToUpperCamelCase;
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
+use std::collections::BTreeSet;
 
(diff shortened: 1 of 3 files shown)


sfackler:
Out of curiosity, what's the use case for doing this? My general preference up to now was to have the code generator parse out the tags and handle all of the special casing up front rather than deferring it to runtime.

sfackler:
Yeah I think the easiest thing for that is to just have conjure-codegen inject a `Content-Encoding: identity` header. Avoids needing to mess with both repos as well.

PR #500, opened 2025-09-26
Add encoding support to clients

sfackler:
## Before this PR
Clients could only send and receive JSON.

## After this PR

==COMMIT_MSG==
The encoding used for request and response bodies from Conjure clients can now be configured. By default, clients now prefer Smile-encoded response bodies over JSON.
==COMMIT_MSG==


conjure-http/src/encoding.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2025 Palantir Technologies, Inc.
```
sfackler, line 1: This is just a copy/paste out of server/encoding.rs

PR #503, opened 2025-10-17
Fix smile serialization of nested values

sfackler:
## Before this PR
We failed to override `is_human_readable` in our nested serializers, so we would incorrectly serialize things like UUIDs whose behavior depends on that flag.

## After this PR

==COMMIT_MSG==
Fixed Smile deserialization of nested UUIDs.
==COMMIT_MSG==

### DIFF
--- a/conjure-serde/src/smile/test.rs
+++ b/conjure-serde/src/smile/test.rs
@@ -171,3 +171,24 @@ fn null_collections() {
         Collections::deserialize(&mut crate::smile::ClientDeserializer::from_slice(smile)).unwrap();
     assert_eq!(expected, actual);
 }
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+struct UuidField {
+    uuid: Uuid,
+}
+
+#[test]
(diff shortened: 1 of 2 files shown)


conjure-serde/src/smile/test.rs
```diff
@@ -171,3 +171,24 @@ fn null_collections() {
...
+}
+
+#[test]
+fn uuid_field() {
+    let smile = b":)\n\x05\xfa\x83uuid\xfd\x90\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xfb";
+
+    let value = UuidField { uuid: Uuid::nil() };
+
+    let actual = crate::smile::to_vec(&value).unwrap();
+    assert_eq!(actual, smile);
```
sfackler: Before the fix, this would serialize as

```
:)\n\x05\xfa\x83uuidc00000000-0000-0000-0000-000000000000\xfb
```

arseny-bogomolov (approved):

PR #553, opened 2026-03-28
Add support for CBOR ser/de to conjure-serde

evanandel:
## Before this PR
conjure-serde only provided support for SMILE binary serialization/deserialization, however CBOR support is required for compatibility in some cases.

## After this PR

==COMMIT_MSG==
Add support for CBOR ser/de to conjure-serde
==COMMIT_MSG==

### DIFF
--- a/conjure-codegen/example-types/example-types.yml
+++ b/conjure-codegen/example-types/example-types.yml
@@ -159,6 +159,15 @@ types:
           safelongs: map<SafeLongAliasExample, ManyFieldExample>
           datetimes: map<DateTimeAliasExample, ManyFieldExample>
           uuids: map<UuidAliasExample, ManyFieldExample>
+      CborMapKeyTests:
+        docs: Test maps with various key types for CBOR Java compatibility
+        fields:
+          stringKeyMap:
+            type: map<string, string>
+            docs: String keys are the standard case
+          uuidKeyMap:
+            type: map<uuid, string>
+            docs: UUID keys get stringified for Java compatibility
       ReservedKeyExample:
         fields:
           package:
--- a/conjure-codegen/example-types/example-types.yml
+++ b/conjure-codegen/example-types/example-types.yml
@@ -159,15 +159,6 @@ types:
           safelongs: map<SafeLongAliasExample, ManyFieldExample>
           datetimes: map<DateTimeAliasExample, ManyFieldExample>
           uuids: map<UuidAliasExample, ManyFieldExample>
-      CborMapKeyTests:
-        docs: Test maps with various key types for CBOR Java compatibility
-        fields:
-          stringKeyMap:
-            type: map<string, string>
-            docs: String keys are the standard case
-          uuidKeyMap:
-            type: map<uuid, string>
-            docs: UUID keys get stringified for Java compatibility
       ReservedKeyExample:
         fields:
           package:
--- a/conjure-serde/Cargo.toml
+++ b/conjure-serde/Cargo.toml
@@ -10,13 +10,13 @@ readme = "../README.md"
 
 [dependencies]
 base64 = "0.22"
+conjure-object = { version = "5.10.0", path = "../conjure-object" }
 serde = "1.0"
 serde_cbor_2 = "0.13"
 serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde-smile = "0.3.1"
 
 [dev-dependencies]
-conjure-object = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
 serde_cbor_2 = "0.13"
--- a/conjure-serde/Cargo.toml
+++ b/conjure-serde/Cargo.toml
@@ -20,4 +20,3 @@ serde-smile = "0.3.1"
 conjure-codegen = { version = "5.10.0", path = "../conjure-codegen", features = ["example-types"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
-serde_cbor_2 = "0.13"
--- a/conjure-serde/Cargo.toml
+++ b/conjure-serde/Cargo.toml
@@ -17,6 +17,7 @@ serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde-smile = "0.3.1"
 
 [dev-dependencies]
+conjure-codegen = { version = "5.10.0", path = "../conjure-codegen", features = ["example-types"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
 serde_cbor_2 = "0.13"
--- a/conjure-serde/Cargo.toml
+++ b/conjure-serde/Cargo.toml
@@ -17,6 +17,5 @@ serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde-smile = "0.3.1"
 
 [dev-dependencies]
-conjure-codegen = { version = "5.10.0", path = "../conjure-codegen", features = ["example-types"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
--- a/conjure-serde/Cargo.toml
+++ b/conjure-serde/Cargo.toml
@@ -11,6 +11,7 @@ readme = "../README.md"
 [dependencies]
 base64 = "0.22"
 serde = "1.0"
+serde_cbor_2 = "0.13"
 serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde-smile = "0.3.1"
 
@@ -18,3 +19,4 @@ serde-smile = "0.3.1"
 conjure-object = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
+serde_cbor_2 = "0.13"
--- a/conjure-serde/src/cbor/cbor_map_key_tests.rs
+++ b/conjure-serde/src/cbor/cbor_map_key_tests.rs
@@ -1,0 +1,61 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_cbor_map_key_bidirectional() {
+        use conjure_codegen::example_types::objects::product::CborMapKeyTests;
+        use conjure_object::Uuid;
+        use std::collections::BTreeMap;
+        let mut string_map = BTreeMap::new();
+        string_map.insert("key1".to_string(), "value1".to_string());
+        string_map.insert("key2".to_string(), "value2".to_string());
+
+        let mut uuid_map = BTreeMap::new();
+        uuid_map.insert(
+            Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000").unwrap(),
+            "first".to_string(),
+        );
+        uuid_map.insert(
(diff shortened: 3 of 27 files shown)


conjure-serde/src/cbor/ser.rs
```diff
@@ -0,0 +1,56 @@
...
+    T: ser::Serialize,
+{
+    serde_cbor_2::to_writer(writer, value)
+}
+
+/// Serializes a `BTreeMap<Uuid, V>` with UUID keys as strings for Java compatibility.
+///
+/// Use this with `#[serde(serialize_with = "conjure_serde::cbor::serialize_uuid_map")]`
+/// on fields with UUID keys to ensure Java can deserialize them.
+pub fn serialize_uuid_map<V, S>(
```
sfackler: This isn't going to work with maps keyed by aliases of Uuids.

conjure-serde/src/cbor/ser.rs
```diff
@@ -0,0 +1,56 @@
...
+    W: Write,
+    T: ser::Serialize,
+{
+    serde_cbor_2::to_writer(writer, value)
+}
+
+/// Serializes a `BTreeMap<Uuid, V>` with UUID keys as strings for Java compatibility.
+///
+/// Use this with `#[serde(serialize_with = "conjure_serde::cbor::serialize_uuid_map")]`
+/// on fields with UUID keys to ensure Java can deserialize them.
```
sfackler: Is the stringification behavior here unique to UUIDs? Integers, booleans, etc are not stringified?

conjure-serde/src/cbor/ser.rs
evanandel: Fixed

conjure-serde/src/cbor/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+//! * Conjure servers reject unknown fields
+//!
+//! This module provides both convenience functions and `Deserializer` types that wrap
+//! `serde_cbor_2` to implement these Conjure-specific behaviors using the behavior composition
+//! pattern shared with JSON and Smile modules.
+//!
+//! Note: There are specific modifications over base serde_cbor_2 to handle compatibility with Java Jackson CBOR
+//!
+//! 1 Java serializes `Map<Integer, String>` with CBOR integer keys, but Rust expected string keys.
+//!   - Added `IntToStringVisitor` in `conjure-serde/src/cbor/de/client.rs` that converts CBOR integers to strings during deserialization.
```
sfackler: This seems wrong to me - there shouldn't be any need for integers to bounce through strings during deserialization.

conjure-serde/src/cbor/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+//! 2. Java serializes `byte[]` as CBOR byte strings (major type 2), but Rust's `Vec<u8>` expected CBOR arrays (major type 4).
```
sfackler: That's expected from serde - byte arrays are represented specially in Rust with e.g. https://docs.rs/serde_bytes/latest/serde_bytes/struct.ByteBuf.html

conjure-serde/src/cbor/ser.rs
```diff
@@ -0,0 +1,85 @@
...
+///
+/// # Example
+/// ```ignore
+/// #[derive(Serialize)]
+/// struct MyType {
+///     #[serde(serialize_with = "conjure_serde::cbor::serialize_map_keys_as_strings")]
+///     uuid_alias_map: BTreeMap<UuidAliasExample, String>,
+/// }
+/// ```
+pub fn serialize_map_keys_as_strings<K, V, S>(
```
sfackler: Why do we need both methods?

Applying these globally at codegen time is going to break Smile.

conjure-serde/src/cbor/mod.rs
evanandel: Ah, true. This was a holdover from when I was still trying to create maximum similarity rather than simply compatibility.

conjure-serde/src/cbor/mod.rs
evanandel: Are you saying this shouldn't need special handling, or just that the handling is expected?

conjure-serde/src/cbor/ser.rs
evanandel: You're right about the duplication. I'll fix that.

For smile generation - as I understand it, it should actually be fine as the smile serialization already enforces string keys, so this behavior is only redundant rather than destructive. I tested serialization and deserialization of a generated conjure object with this flag in both cbor and smile and it worked for both. Is there something I'm overlooking?

conjure-serde/src/cbor/mod.rs
sfackler: It's expected that this will end up serializing as a seq of u8:

```rust
#[derive(Serialize)]
struct Foo {
    x: Vec<u8>,
}
```

while this will serialize as a byte sequence:

```rust
#[derive(Serialize)]
struct Foo {
    x: serde_bytes::ByteBuf, // or bytes::Bytes
}
```

evanandel: Ah, I get it now. Thank you. Removed the unnecessary handling.

conjure-serde/src/cbor/ser.rs
evanandel: Java is fine deserializing the int keys as ints, even if that's not how it serializes. It's only the UUIDs that it _needs_ serialized in its preferred format.

evanandel:
I _think_ this should be all set. I also included the build.gradle fix for the broken publish task.

evanandel:
Note: the changes to codegen introduce a hidden dependency on the changes in serde. Unless there's a way to update the dependency in codegen in advance of a new serde release being triggered, this PR is unsafe as written. I'm going to leave it as-is for review purposes. But once I have sign-off (formal or informal) on the contents I'm going to pull out the codegen changes into a separate PR that can depend on a release triggered by this one.

conjure-serde/src/cbor/de/client.rs
```diff
@@ -0,0 +1,174 @@
...
+        false
+    }
+}
+
+pub enum ValueBehavior {}
+
+impl Behavior for ValueBehavior {
+    type KeyBehavior = KeyBehavior;
+
+    fn deserialize_byte_buf<'de, D, V>(de: D, visitor: V) -> Result<V::Value, D::Error>
```
sfackler: Isn't this override a no-op?

conjure-serde/src/cbor/de/client.rs
```diff
@@ -0,0 +1,174 @@
...
+
+impl Behavior for KeyBehavior {
+    type KeyBehavior = Self;
+
+    fn deserialize_bytes<'de, D, V>(de: D, visitor: V) -> Result<V::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+        V: Visitor<'de>,
+    {
+        // UUID keys: accept both string format (from Java) and binary format (backward compat)
```
sfackler: What back compat do we need to preserve here?

conjure-serde/src/cbor/ser.rs
sfackler: This should go in `conjure-object::private` - that's where we already have some similar things.

conjure-serde/Cargo.toml
```diff
@@ -10,11 +10,14 @@ readme = "../README.md"
...
+serde_cbor_2 = "0.13"
 serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde-smile = "0.3.1"
 
 [dev-dependencies]
-conjure-object = "1.1.0"
+conjure-codegen = { version = "5.10.0", path = "../conjure-codegen", features = ["example-types"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11"
+serde_cbor_2 = "0.13"
```
sfackler: No need to add this as a dev dependency if it's already a runtime dependency.

conjure-serde/src/cbor/cbor_map_key_tests.rs
```diff
@@ -0,0 +1,68 @@
+#[cfg(test)]
+mod tests {
```
sfackler, line 2: No need for this extra level of module.

conjure-serde/Cargo.toml
```diff
@@ -10,11 +10,14 @@ readme = "../README.md"
...
 base64 = "0.22"
+conjure-object = { version = "5.10.0", path = "../conjure-object" }
 serde = "1.0"
+serde_cbor_2 = "0.13"
 serde_json = { version = "1.0", features = ["float_roundtrip"] }
 serde-smile = "0.3.1"
 
 [dev-dependencies]
-conjure-object = "1.1.0"
+conjure-codegen = { version = "5.10.0", path = "../conjure-codegen", features = ["example-types"] }
```
sfackler: Why do we need this dependency?

evanandel: It's used in the test_cbor_map_key_bidirectional test. If you don't like the dev dep I can just kill that test.

conjure-serde/src/cbor/de/client.rs
evanandel: You're right here and below; this was left over from the earlier needlessly fiddly byte handling.

conjure-serde/Cargo.toml
sfackler: Ah - tests that depend on conjure-generated objects can more easily go in the conjure-test crate.

evanandel: That makes sense. Fixed.

conjure-codegen/example-types/example-types.yml
```diff
@@ -159,6 +159,15 @@ types:
           safelongs: map<SafeLongAliasExample, ManyFieldExample>
           datetimes: map<DateTimeAliasExample, ManyFieldExample>
           uuids: map<UuidAliasExample, ManyFieldExample>
+      CborMapKeyTests:
```
sfackler: conjure-test has its own Conjure definitions - this case should be moved over to that.

conjure-serde/src/cbor/de/client.rs
sfackler: Seems like this is still here?

conjure-serde/src/cbor/ser.rs
sfackler: This still needs to move.

conjure-serde/src/cbor/de/client.rs
evanandel: Huh, I swear earlier my tests were failing if I dropped this, but that's no longer the case. I must have mixed something up. Removed.

conjure-serde/src/cbor/ser.rs
evanandel: I'm sorry, I somehow the object::private comment last week. Moved.

conjure-test/Cargo.toml
```diff
@@ -15,11 +15,14 @@ base64 = "0.22"
 bytes = "1.0"
 conjure-macros = { path = "../conjure-macros" }
 conjure-serde = { path = "../conjure-serde" }
+conjure-codegen = { path = "../conjure-codegen", features = ["example-types"]  }
```
sfackler: I don't think this is needed.

sfackler (approved):

PR #559, opened 2026-05-28
Log Safety

aiverson-pt:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/conjure-codegen/src/enums.rs
+++ b/conjure-codegen/src/enums.rs
@@ -20,7 +20,14 @@ use crate::types::objects::EnumDefinition;
 pub fn generate(ctx: &Context, def: &EnumDefinition) -> TokenStream {
     let enum_ = generate_enum(ctx, def);
     let unknown = generate_unknown(ctx, def);
-    let log_safe = ctx.log_safe_impl(def.type_name());
+    let log_safe = if ctx.is_safe_type(def.type_name()) {
+        let name = ctx.type_name(def.type_name().name());
+        quote! {
+            impl conjure_object::log_safety::LogSafe for #name {}
+        }
+    } else {
+        quote!()
+    };
 
     quote! {
         // https://github.com/serde-rs/serde/issues/2195
--- a/conjure-codegen/src/enums.rs
+++ b/conjure-codegen/src/enums.rs
@@ -20,14 +20,6 @@ use crate::types::objects::EnumDefinition;
 pub fn generate(ctx: &Context, def: &EnumDefinition) -> TokenStream {
     let enum_ = generate_enum(ctx, def);
     let unknown = generate_unknown(ctx, def);
-    let log_safe = if ctx.is_safe_type(def.type_name()) {
-        let name = ctx.type_name(def.type_name().name());
-        quote! {
-            impl conjure_object::log_safety::LogSafe for #name {}
-        }
-    } else {
-        quote!()
-    };
 
     quote! {
         // https://github.com/serde-rs/serde/issues/2195
@@ -37,7 +29,6 @@ pub fn generate(ctx: &Context, def: &EnumDefinition) -> TokenStream {
 
         #enum_
         #unknown
-        #log_safe
     }
 }
 
@@ -48,6 +39,24 @@ fn generate_enum(ctx: &Context, def: &EnumDefinition) -> TokenStream {
     let ok = ctx.ok_ident(def.type_name());
     let err = ctx.err_ident(def.type_name());
     let unknown = unknown(ctx, def);
+
+    let mut derives = vec![
+        "Debug",
+        "Clone",
+        "PartialEq",
+        "Eq",
+        "PartialOrd",
+        "Ord",
+        "Hash",
+        "conjure_object::serde::Deserialize",
+        "conjure_object::serde::Serialize",
+    ];
+
+    if ctx.is_safe_type(def.type_name()) {
+        derives.push("conjure_object::log_safety::derive::LogSafe");
+    }
+
+    let derives = derives.iter().map(|s| s.parse::<TokenStream>().unwrap());
 
     let variants = def.values().iter().map(|v| {
         let docs = ctx.docs(v.docs());
@@ -110,17 +119,7 @@ fn generate_enum(ctx: &Context, def: &EnumDefinition) -> TokenStream {
 
     quote! {
         #root_docs
-        #[derive(
-            Debug,
-            Clone,
-            PartialEq,
-            Eq,
-            PartialOrd,
-            Ord,
-            Hash,
-            conjure_object::serde::Deserialize,
-            conjure_object::serde::Serialize,
-        )]
+        #[derive(#(#derives),*)]
         #[serde(crate = "conjure_object::serde")]
         pub enum #name {
             #(#variants)*
--- a/conjure-codegen/src/enums.rs
+++ b/conjure-codegen/src/enums.rs
@@ -20,6 +20,7 @@ use crate::types::objects::EnumDefinition;
 pub fn generate(ctx: &Context, def: &EnumDefinition) -> TokenStream {
     let enum_ = generate_enum(ctx, def);
     let unknown = generate_unknown(ctx, def);
+    let log_safe = ctx.log_safe_impl(def.type_name());
 
     quote! {
         // https://github.com/serde-rs/serde/issues/2195
@@ -29,6 +30,7 @@ pub fn generate(ctx: &Context, def: &EnumDefinition) -> TokenStream {
 
         #enum_
         #unknown
+        #log_safe
     }
 }
 
--- a/conjure-codegen/src/enums.rs
+++ b/conjure-codegen/src/enums.rs
@@ -204,6 +204,7 @@ fn generate_unknown(ctx: &Context, def: &EnumDefinition) -> TokenStream {
             Hash,
             conjure_object::serde::Deserialize,
             conjure_object::serde::Serialize,
+            conjure_object::log_safety::derive::LogSafe,
         )]
         #[serde(crate = "conjure_object::serde", transparent)]
         pub struct #unknown(conjure_object::private::Variant);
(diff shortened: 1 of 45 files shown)


conjure-codegen/src/enums.rs
```diff
@@ -20,6 +20,14 @@ use crate::types::objects::EnumDefinition;
 pub fn generate(ctx: &Context, def: &EnumDefinition) -> TokenStream {
     let enum_ = generate_enum(ctx, def);
     let unknown = generate_unknown(ctx, def);
+    let log_safe = if ctx.is_safe_type(def.type_name()) {
+        let name = ctx.type_name(def.type_name().name());
+        quote! {
+            impl conjure_object::log_safety::LogSafe for #name {}
```
sfackler: Why aren't we deriving for enums?

conjure-macros/src/lib.rs
```diff
@@ -424,6 +425,14 @@ pub fn derive_with(input: proc_macro::TokenStream) -> TokenStream {
     derive_with::generate(input)
 }
 
+/// Marks the annotated type as safe to log.
+///
+/// The `conjure_object` crate must be in scope, and its `log-safety` feature must be enabled.
```
sfackler: It might simplify things a bit to always have LogSafe defined, and just gate MaybeLogSafe on the log-safety feature.

conjure-macros/src/log_safety.rs
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, parse_quote, DeriveInput};
+
+pub fn generate(input: TokenStream) -> TokenStream {
```
sfackler, line 22: I think we'll want to have this assert that all fields are logsafe (maybe overridable by an attribute). The standard library's `#[derive(Eq)]` does this as an example.

conjure-codegen/src/enums.rs
aiverson-pt: The enum code gen has a hardcoded derives (L:133).

conjure-object/src/log_safety.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::collections::{BTreeMap, HashMap};
+
+/// Marker trait for types whose serialized representation is safe to include
+/// in logs and error parameters.
+pub trait LogSafe {}
+
+/// Wrapper struct for known-safe data.
+#[derive(serde::Serialize)]
+pub struct AssertLogSafe<T: serde::Serialize>(pub T);
+impl<T: serde::Serialize> LogSafe for AssertLogSafe<T> {}
```
sfackler: It'd be more typical to omit the Serialize bounds here - Rust APIs typically defer the bounds to the places they're actually needed, even if that allows you to e.g. construct a useless value.

conjure-macros/src/lib.rs
aiverson-pt: This is an outdated comment from my first approach. We do take the feature gated `MaybeLogSafe` approach now. Addressed.

conjure-object/src/log_safety.rs
```diff
@@ -0,0 +1,62 @@
...
+//!
+//! Enable the `log-safety` cargo feature to opt in.
+
+use std::collections::{BTreeMap, HashMap};
+
+/// Marker trait for types whose serialized representation is safe to include
+/// in logs and error parameters.
+pub trait LogSafe {}
+
+/// Wrapper struct for known-safe data.
```
sfackler, line 22: We should have a big warning here that this is an escape hatch that it's on the developer to not misuse.

conjure-object/src/log_safety.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Marker bound that resolves to `LogSafe` when `log-safety` is enabled,
+/// or is satisfied by any type when it's not.
+#[cfg(not(feature = "log-safety"))]
+pub trait MaybeLogSafe {}
+
+#[cfg(not(feature = "log-safety"))]
+impl<T> MaybeLogSafe for T {}
+
+// conjure-object types that are safe
+impl LogSafe for crate::ResourceIdentifier {}
```
sfackler: Nit: I'd put this one in the rid module

conjure-object/src/log_safety.rs
```diff
@@ -0,0 +1,62 @@
...
+
+// containers
+impl<T: LogSafe + ?Sized> LogSafe for &T {}
+impl<T: LogSafe + ?Sized> LogSafe for &mut T {}
+impl<T: LogSafe + ?Sized> LogSafe for Box<T> {}
+impl<T: LogSafe> LogSafe for Option<T> {}
+impl<T: LogSafe> LogSafe for Vec<T> {}
+impl<T: LogSafe, const N: usize> LogSafe for [T; N] {}
+impl<K: LogSafe, V: LogSafe> LogSafe for BTreeMap<K, V> {}
+impl<K: LogSafe, V: LogSafe> LogSafe for HashMap<K, V> {}
```
sfackler: There's probably some more std types to add here - BTree/HashSet, Duration, VecDeque, Arc, Rc, etc

conjure-error/Cargo.toml
```diff
@@ -8,6 +8,9 @@ description = "Runtime support for generated Conjure errors"
 repository = "https://github.com/palantir/conjure-rust"
 readme = "../README.md"
 
+[features]
+log-safety = ["conjure-object/log-safety"]
```
sfackler: I don't think we actually need this feature.

conjure-error/src/error.rs
```diff
@@ -280,7 +281,7 @@ impl Error {
     /// Panics if the value fails to serialize.
```
sfackler, line 286: Probably worth mentioning the bound and AssertLogSafe in here?

conjure-codegen/src/enums.rs
sfackler: We should just change that I think

conjure-macros/src/log_safety.rs
aiverson-pt: Hmm... Not 100% sure about this. We definitely don't want to implement LogSafe by default on things like `String`. But if my code gen looks like `pub struct MimeType(pub String);`, we'd have to go the attribute route.

The thing I dislike about that is that our codegen needs to track two things now, the derive and the attribute.

sfackler: Yeah I think we'll want `#[derive(LogSafe)] pub struct MimeType(#[assert_log_safe] pub String);`

conjure-codegen/src/enums.rs
```diff
@@ -40,6 +40,24 @@ fn generate_enum(ctx: &Context, def: &EnumDefinition) -> TokenStream {
...
+        "Eq",
+        "PartialOrd",
+        "Ord",
+        "Hash",
+        "conjure_object::serde::Deserialize",
+        "conjure_object::serde::Serialize",
+    ];
+
+    if ctx.is_safe_type(def.type_name()) {
+        derives.push("conjure_object::log_safety::derive::LogSafe");
```
sfackler: Should this be unconditional? Can you define unsafe/donotlog enums in conjure?

conjure-http/src/server/mod.rs
```diff
@@ -1043,7 +1043,7 @@ where
     if remaining > 0 {
         return Err(
             Error::service_safe("expected at most 1 parameter", InvalidArgument::new())
-                .with_safe_param("actual", remaining + 1),
+                .with_safe_param("actual", AssertLogSafe(remaining + 1)),
```
sfackler, line 1046: This is kind of annoying... I wonder how java log safety handles it?

conjure-macros/src/log_safety.rs
```diff
@@ -0,0 +1,66 @@
...
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = &input.ident;
+
+    let field_types = collect_field_types(&input.data);
+
+    let mut generics = input.generics.clone();
+    {
+        let where_clause = generics.make_where_clause();
+        for ty in &field_types {
```
sfackler, line 31: I think I'd rather have this fail to compile than generate an impl that silently does nothing.

conjure-object/src/log_safety.rs
```diff
@@ -0,0 +1,72 @@
...
+impl<T: LogSafe> LogSafe for Vec<T> {}
+impl<T: LogSafe> LogSafe for VecDeque<T> {}
+impl<T: LogSafe> LogSafe for std::collections::LinkedList<T> {}
+impl<T: LogSafe> LogSafe for [T] {}
+impl<T: LogSafe, const N: usize> LogSafe for [T; N] {}
+impl<K: LogSafe, V: LogSafe> LogSafe for BTreeMap<K, V> {}
+impl<K: LogSafe, V: LogSafe, S> LogSafe for HashMap<K, V, S> {}
+impl<T: LogSafe> LogSafe for std::collections::BTreeSet<T> {}
+impl<T: LogSafe, S> LogSafe for std::collections::HashSet<T, S> {}
+impl<T: LogSafe> LogSafe for std::collections::BinaryHeap<T> {}
```
sfackler, line 65: We should be consistent about importing all these types

conjure-macros/src/log_safety.rs
sfackler: Nevermind! My assumption about how these clauses would behave was wrong.

conjure-http/src/server/mod.rs
aiverson-pt: We could mark `usize` as log safe... That is closest to java log safety primatives.

sfackler:
:+1:

PR #590, opened 2026-08-26
Add support for setting a custom backtrace provider

evanandel:
## Before this PR
`backtrace::Backtrace` doesn't support platforms like `wasm`. They are thus either stuck with no backtraces, or need to call `with_custom_safe_backtrace` on every error (at the point of creation to get the correct trace).

## After this PR

==COMMIT_MSG==
Allows the setting of a custom backtrace provider on platforms without `backtrace::Backtrace` support.
==COMMIT_MSG==
An example provider would be constructing a string from `js_sys`'s stack view.

### DIFF
--- a/conjure-error/src/error.rs
+++ b/conjure-error/src/error.rs
@@ -19,6 +19,7 @@ use std::borrow::Cow;
 use std::collections::hash_map::{self, HashMap};
 use std::fmt;
 use std::ops::Index;
+use std::sync::OnceLock;
 use std::time::Duration;
 use std::{backtrace, error};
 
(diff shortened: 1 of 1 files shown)


conjure-error/src/error.rs
```diff
@@ -403,6 +404,17 @@ impl<'a> Iterator for ParamsIter<'a> {
     }
 }
 
+static BACKTRACE_PROVIDER: OnceLock<fn() -> String> = OnceLock::new();
+
+/// Overrides backtrace capture. The returned string lands verbatim in the
+/// `stacktrace` field of service.1 records and is treated as safe-to-log. Used for platforms
+/// like wasm with that backtrace::Backtrace doesn't support.
+pub fn set_safe_custom_backtrace_provider(provider: fn() -> String) -> Result<(), Error> {
```
sfackler: Do you think it's worth accepting a closure here instead of a bare fn? I could imagine a custom backtrace implementation might want to have some state for e.g. caches.

evanandel: Good idea. The cross-thread access means it's got to be a `static closure which may restrict a bit, but still provides more generality.

sfackler (approved):

